PR #3, opened 2017-07-21
Week2 React Homework Essam

essamalsaloum:


todo-app/src/App.js
```diff
@@ -1,21 +1,107 @@
...
+      todosList[index] = editToDo
+      this.setState({
+        todosList
+        });
+    }
+
+    render() {
+      let todoList=this.state.todos
+      todoList= todoList.map(function(item,index){
+        return( <List  description={item.description}
```
spirosikmd: I am not sure why you used another component here `List` that just renders a `ListItem`?  If you check your react dev tools with chrome, this code renders two `List` components.

screen shot 2017-07-21 at 21 57 14

What you probably wanted to do here is to render one `List` component and pass as a prop the `todoList`. The `List` component then would render in a map the `ListItem` components. So this code would end up being something like

```javascript
<List
  todos={todoList}
  onDelete={this.onDelete}
  handleCheckBoxChange={this.handleCheckBoxChange}
  save={this.save}
/>
```

Does this make sense? Do you think this is a better approach?

todo-app/src/App.js
```diff
@@ -1,21 +1,107 @@
...
+        }
+
+    }
+
+    handleSubmit(event){
+      event.preventDefault()
+      const {newToDo} = this.state  
+      const todos = [...this.state.todos, newToDo]
+      this.setState({todos})
+      document.getElementById("addItem").reset();
```
spirosikmd: We should avoid using `document.getElementById` and try to use what React offers us in terms of event handling, data manipulation and binding, and state. I guess you are using this line of code to clear the value of the description input. However, if you turn `<input type="text"  onChange={this.handleChange.bind(this)}/>` into a [controller component](https://facebook.github.io/react/docs/forms.html#controlled-components)

```html
<input type="text" value={this.state.newTodo.description} onChange={this.handleChange.bind(this)}/>
```

you wouldn't need to do this and just resetting the state with

```javascript
this.setState({
  newToDo: {
    description: '',
    done: false
  }
});
```

would do the job! Does this make sense?

todo-app/src/App.js
```diff
@@ -1,21 +1,107 @@
...
         </div>
-        <p className="App-intro">
-          To get started, edit <code>src/App.js</code> and save to reload.
-        </p>
-      </div>
-    );
+      )
+    }
+
+    onDelete= (index) => {
```
spirosikmd: We should avoid mutating objects and arrays and use functions that help us write immutable code. Immutable code leads to fewer bugs and more performant code. In this case you could use [`slice`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/slice) method to copy over the todos array.

todo-app/src/App.js
```diff
@@ -1,21 +1,107 @@
...
+      const todosList = this.state.todos
+      todosList.splice(index, 1)
+      this.setState({todosList})
+    }
+
+
+
+    handleCheckBoxChange = (index) => {
```
spirosikmd: Same here in terms of writing this code with immutability in mind. You can use the slice method, and possibly [`Object.assign`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) to not mutate the todo item, which you do now by using the `index` and setting the `done` property.

todo-app/src/App.js
```diff
@@ -1,21 +1,107 @@
...
+      const todosList = this.state.todos
+      todosList[index].done = !todosList[index].done;
+      this.setState({
+        todosList
+      });
+    }
+
+    save = (editToDo,index) => {
```
spirosikmd: Same comment here regarding immutability 😄

todo-app/src/list.js
```diff
@@ -0,0 +1,23 @@
...
+import ListItem from './ListItem'
+import './App.css';
+
+ export default class List extends React.Component {
+
+render() {
+
+    return (
+      <div className='todo-items'>
+        <ListItem description={this.props.description}
```
spirosikmd: If you agree with my previous comment about `List` and `ListItem`, and assuming `List` component would receive an array of `todos` as a prop, then here you would write something like

```javascript
const todos = this.props.todos;
{todos.map((item, index) => {
  return (
    <ListItem
      key={index}
      index={index}
      // the rest of the properties here
    />
  );
}}
```

todo-app/src/ListItem.js
```diff
@@ -0,0 +1,72 @@
...
+
+handelsave(event){
+  event.preventDefault()
+  this.props.save(this.state.editToDo, this.props.index)
+  this.handelTextClick()
+}
+
+handelTextClick(){
+  this.setState(prevState => ({
+      showEdit: !prevState.showEdit
```
spirosikmd: Nice 👍

But this can lead to confusion. Whenever we want to change the edit state of the `ListItem` component we call `handleTextClick` and we assume that we get the correct result. Would it make it more clear to rename the function according to its scope, pass the value you want directly to the function

```javascript
setShowEdit (showEdit) {
  this.setState({ showEdit }); // note here we are using the object property shorthand
});
```

and use it like `this.setShowEdit(true)` or `this.setShowEdit(false)`? What do you think?

todo-app/src/ListItem.js
```diff
@@ -0,0 +1,72 @@
...
+      showEdit: true,
+      editToDo: {
+        description: props.description,
+        done: false
+      }
+    }
+
+  }
+
+  Edit(showEdit) {
```
spirosikmd: Why did you use a capital letter here? This is just a function of the `ListItem` component. If you meant for this to be another component, then you should put it in another file like `Edit.js` and import it and use it here.

todo-app/src/ListItem.js
```diff
@@ -0,0 +1,72 @@
...
+    if (!showEdit) {
+      return (
+        <div className='item-name'>
+        <form id="addItem"  onSubmit={this.handelsave.bind(this)}>
+          <input type="text" value={this.state.editToDo.description} onChange={this.handleTextChange.bind(this)}/>
+          <input type="submit" value="Save" />
+          <button type="button" onClick={this.handelTextClick.bind(this)} > Cancel </button>
```
spirosikmd: Maybe it would be more descriptive to name this `this.handleCancel`? What do you think?

PR #4, opened 2017-07-22
Homework for week 2 (todo app)

yasershomaf:
Homework for week 2 is done


todo-app/src/App.js
```diff
@@ -1,21 +1,174 @@
...
+              done: false,
+              toDoId: this.state.nextId,
+              action: "Edit"
+            }),
+          nextId: this.state.nextId + 1,
+        });
+      }
+    }
+  }
+  handleControllerBTN = () => {
```
spirosikmd: The function name we use here, doesn't clearly indicate what this function does. Maybe it would be easier to write two functions, for example 

```javascript
setDeleteAllIsClicked = (deletAllIsClicked) => {
    this.setState({ deletAllIsClicked });
}
```

and call it like `this.setDeleteAllIsClicked(true)`. Does this make sense? What do you think?

todo-app/src/App.js
```diff
@@ -1,21 +1,174 @@
...
+    this.setState({
+      deletAllIsClicked: !this.state.deletAllIsClicked
+    });
+  }
+  findTodoIndex = (prefix, id) => {
```
spirosikmd: If we would use only the todo object id and not the input element ids, we wouldn't need to pass around the prefix. In this todo app, we can accomplish everything we need without using the input element ids. Just pass around `todo.id`. How would you change this code to accomplish that?

todo-app/src/App.js
yasershomaf: I dont think that passing an argument to this function is necessary or even proper, because the only purpose of this function is to invert the state of this.setState.deletAllIsClicked (this.setState.clearIsClicked in the new version because I've changed the name).
This function decides the mission of the first button (is it gonna be add or clear?). That's why I gave it this name (ControllerBTN). But I think that (as you said) the name still doesn't clearly indicate what this function does. I'll try to find a better one

PR #5, opened 2017-07-22
Master

Ahalhowidi:
Hameed homework week2

### DIFF
--- a/todo-app/src/App.js
+++ b/todo-app/src/App.js
@@ -25,14 +25,7 @@ import Todo from './Todo';
   }
     
   
-    onAdd = () => {
-
-    const description = prompt("Add the description ?")
-
-    const newToDo = {description};
-    const todosList = [...this.state.todos, newToDo];
-    this.setState({todos:todosList})
-    }
+   
   onDelete= (index) => {
     const todosList = [...this.state.todos];
     todosList.splice(index, 1);
--- a/todo-app/src/App.js
+++ b/todo-app/src/App.js
@@ -1,21 +1,63 @@
-import React, { Component } from 'react';
+import React from 'react';
 import logo from './logo.svg';
 import './App.css';
+import Todo from './Todo';
 
-class App extends Component {
-  render() {
-    return (
-      <div className="App">
-        <div className="App-header">
-          <img src={logo} className="App-logo" alt="logo" />
-          <h2>Welcome to React</h2>
-        </div>
-        <p className="App-intro">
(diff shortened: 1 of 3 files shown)


todo-app/src/App.js
```diff
@@ -1,21 +1,56 @@
...
+      return (
+        <div className="App">
+          <div className="App-header">
+            <img src={logo} className="App-logo" alt="logo" />
+            <h2>To Do List</h2>
+          </div>
+          <div className= "App-Body">
+          {todoList}
+          </div>
+           <input ref={input => this.name = input} />
```
spirosikmd: Although there are special cases where `ref` can be used, in general we should [avoid using it](https://facebook.github.io/react/docs/refs-and-the-dom.html#dont-overuse-refs). To get and set the value of this input element, we should better use a [controlled component](https://facebook.github.io/react/docs/forms.html#controlled-components).

todo-app/src/App.js
```diff
@@ -1,21 +1,56 @@
...
+               }
+            ]
+        };
+        
+    }
+    
+    handleChange = (e) => {
+    const description =  this.name.value;
+    const newToDo = {description};
+    const todosList = [...this.state.todos, newToDo];
```
spirosikmd: Immutability nice!

todo-app/src/App.js
```diff
@@ -1,21 +1,56 @@
...
+   
+  onDelete= (index) => {
+    const todosList = [...this.state.todos];
+    todosList.splice(index, 1);
+    this.setState({todos:todosList});
+  }
+    render() {
+      let todoList=this.state.todos;
+      todoList= todoList.map(function(item,index){
+        return( <Todo  description={item.description} key={index} index={index} onDelete={this.onDelete}/>);
```
spirosikmd: You are not passing `done` as prop and that is why it is not rendered in your `Todo` component with `{this.props.done} ...`!

todo-app/src/App.js
```diff
@@ -1,21 +1,56 @@
...
+  
+   
+  onDelete= (index) => {
+    const todosList = [...this.state.todos];
+    todosList.splice(index, 1);
+    this.setState({todos:todosList});
+  }
+    render() {
+      let todoList=this.state.todos;
+      todoList= todoList.map(function(item,index){
```
spirosikmd: Although using `function (item, index){}.bind(this)` works, doesn't it become more readable when using an arrow function, for example `(item, index) => {}`? What do you think?

PR #6, opened 2017-07-24
Problem with deleting items

nos111:
Hi Spyros,

The id shouldn't be changing but the console log is printing a different id when you remove an item which is weird.

Could you please check this?

Thanks alot


todo-app/src/components/TasksForm.js
```diff
@@ -0,0 +1,30 @@
...
+import TaskItem from './TaskItem'
+
+export default class TasksFrom extends Component {
+    render() {
+        const todos = this.props.todos
+        return (
+            <div>
+                {todos.map((task,index) => {
+                    return <TaskItem 
+                        key={index}
```
spirosikmd: The problem with deleting tasks is in this line @nos111! You are using the index as key, and as we mentioned during the last class, using the index might cause unexpected behaviour. React doesn't know that the item that removed with index 1 was different than the new item that is rendered (after deletion) at index 1.

What we want is to use the `task.id` as key and deleting items should work as expected. If you use chrome browser and the react devtools extension, then this is what we should se

screen shot 2017-07-24 at 20 05 11

Notice the keys are now 1, 2, 3, 4 and not 0, 1, 2, 3 when you use the index as key 😄

PR #7, opened 2017-07-25
Completed todo-app

nos111:
Hi Spyros,

I have completed my todo-app.

Could you please review it and leave your comments :)

Thanks alot


todo-app/src/App.js
```diff
@@ -1,20 +1,135 @@
...
+    return item.id === id
+  })
+  const todo = this.state.todos[index]
+  const newTodo = Object.assign({}, todo, {
+    description: newDescription
+  })
+  this.setState({
+    todos: this.state.todos.slice(0, index).concat(newTodo, this.state.todos.slice(index + 1))
+  })
+  console.log(index)
```
spirosikmd: You can remove these logs here I guess 😄

todo-app/src/App.js
```diff
@@ -1,20 +1,135 @@
...
+  console.log(newDescription + id)
+}
+
+handleDelete(newId){
```
spirosikmd: Why are you calling this `newId` here? Is there any specific reason? Although this function works, how would you write it simpler? Hint you can use `.splice` without mutating the original todos array!

todo-app/src/App.js
```diff
@@ -1,20 +1,135 @@
...
+  console.log(this.state.todos)
+
+}
+
+handleCompletedTask(newId){
+  console.log(newId)
+  const index = this.state.todos.findIndex((item) => {
+    return item.id === newId}
+  )
+  const completed = this.state.todos[index].done ? false : true;
```
spirosikmd: Although this is eventually consistent and it works, you are not using the value directly that is coming from the onChange event of the checkbox, but you are reading the one from the state. You might get yourself in trouble doing that. How would you use the value that is coming from the onChange event from the checkbox?

todo-app/src/components/AddTodo.js
```diff
@@ -0,0 +1,44 @@
...
+        this.props.handleNewTaskSubmit(this.state.textValue)
+        this.setState({
+            textValue: ''
+        }
+        )
+    }
+
+    render(){
+        return(
+            <form onSubmit={this.handleLocalSubmit}>
```
spirosikmd: Very nice that you used a form element here 👍

todo-app/src/components/TaskItem.js
```diff
@@ -0,0 +1,45 @@
...
+        e.preventDefault()
+        this.props.handleSubmit(this.state.textValue)
+    }
+    render(){
+        const {index, task} = this.props
+        return(
+            <form onSubmit={this.handleLocalSubmit}>
+            <input id='completed' 
+            type='checkbox' 
+            defaultChecked={task.done}
```
spirosikmd: Instead of using `defaultChecked` and `onClick` here you should turn this into a controller component with `checked` and `onChange`. What do you think?

todo-app/src/components/TaskItem.js
```diff
@@ -0,0 +1,45 @@
...
+            <input id='completed' 
+            type='checkbox' 
+            defaultChecked={task.done} 
+            onClick={() => this.props.handleCompletedTask(task.id)} 
+            />
+            <label>{index + 1}</label>
+            <input id='task' 
+                    type='text' 
+                    required
+                    defaultValue={task.description}
```
spirosikmd: Do you need the `defaultValue` here? Wouldn't just using the `this.state.textValue` as `value` work for you?

todo-app/src/App.js
```diff
@@ -1,20 +1,135 @@
 import React, { Component } from 'react';
 import logo from './logo.svg';
 import './App.css';
+import todo from './todo.json';
```
spirosikmd: This is not used in this file, so maybe you can remove it?

todo-app/src/components/TasksForm.js
```diff
@@ -0,0 +1,31 @@
+import React, { Component } from 'react'
+import PropTypes  from 'prop-types'
```
spirosikmd: This is not used in this file, so maybe you can remove it?

nos111: Yes I shall do that :)

todo-app/src/components/TaskItem.js
nos111: I don't understand the difference between the default value and the value. I think they are the same

todo-app/src/components/TaskItem.js
nos111: I think this is easier. This way I only have to change the state and it will change automatically without much work for me :P

todo-app/src/components/AddTodo.js
nos111: Thank you :)

todo-app/src/App.js
nos111: Yes, Because I'm keeping the change local until the user press enter (submit the form) than I will save what the onChange function has captured as my new description.

todo-app/src/App.js
nos111: newId is here refering to the ID but I needed to give it a different name for the extinguish.
Splice will return the deleted item, which will leave me with only the item I wanted to delete instead of the array without the item I deleted

PR #8, opened 2017-10-20
Todo App Homework 

AlexanderSaid:


spirosikmd:
Hey @AlexanderSaid! I see many more changes unrelated to your homework in this PR. This makes it a bit more difficult to review the homework..

todo-app/src/TasksDB.js
```diff
@@ -0,0 +1,17 @@
+export default [
+    {
+        text: "* Get out of bed",
+        done: "(done)"
```
spirosikmd: For the `done` property you could have used a boolean `true` or `false` value.

todo-app/src/Tasks.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import Description from './Description';
+import Done from './Done';
+import './tasks.css';
+
+export default class Tasks extends React.Component {
```
spirosikmd: Why did you use plural `Tasks` here? Why not just `Task`? This component receives a task object right? What do you think?

todo-app/src/App.js
```diff
@@ -1,13 +1,26 @@
...
 
-class App extends Component {
+export default class App extends React.Component {
   render() {
     return (
       <div>
-        Hello World!
+        <h1>ToDo List:</h1>
+        <div>
+          {task.map((task, index) => {
```
spirosikmd: You could have imported this `task` as `tasks`, e.g. `import tasks from './TasksDB'`, so we avoid using the same variable name inside map and for this list.

todo-app/src/Tasks.js
AlexanderSaid: Ya you are right, I need to improve naming skills

todo-app/src/TasksDB.js
AlexanderSaid: I'll try it today

AlexanderSaid:
@spirosikmd Hey, 
Thanks for the review, I'll work on your notes.
I think that you can ignore some commits because I was trying things

todo-app/src/Status.js
```diff
@@ -1,11 +1,26 @@
...
         return (
-            <div>
-                {this.props.done}
-            </div>
-        )
-    }
+          <label>
+          <input type="checkbox"
+            checked={this.state.isDone}
+            onClick={this.click}
```
spirosikmd: You should use `onChange` here, as it describes better what we are trying to accomplish. We want to update the `checked` we get from the state, when the actual HTML input element `checked` property changes. Does this make sense?

todo-app/src/Status.js
```diff
@@ -1,11 +1,26 @@
 import React from 'react';
 
 export default class Status extends React.Component{
-    render() {
+    constructor(props) {
+        super(props);
+        this.state = {isDone: false};
```
spirosikmd: If you add an `isDone` `boolean` property to each item in `todo-app/src/TasksDB.js`, then you can pass it as prop to `Status` component and set the state.

```js
this.state = {isDone: this.props.isDone};
```

PR #9, opened 2017-10-20
homework added

2sma3il1991juve:


todo-app/src/App.js
```diff
@@ -1,13 +1,24 @@
...
+import React from 'react';
+import TodoList from './TodoList';
+import Todos from './Todos';
 
-class App extends Component {
+export default class App extends React.Component {
+  renderTodos(Todos) {
+    return Todos.map((todo, index) => {
+      return (
+        <TodoList
```
spirosikmd: A better naming for this component as it accepts a single `todo` object is `Todo`? What do you think?

todo-app/src/TodoItem.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+export default class TodoItem extends React.Component {
+  render() {
+    return (
+      <div><ul>
```
spirosikmd: If each `TodoItem` renders a `ul` then for three items we will have three `ul` elements, which is not very nice, right? Ideally we want to have one `ul` and each item should render an `li` element?

todo-app/src/App.js
```diff
@@ -1,13 +1,21 @@
-import React, { Component } from 'react';
+import React from 'react';
+import TodoList from './TodoList';
+import Todos from './Todos';
 
-class App extends Component {
+export default class App extends React.Component {
+  renderTodos(Todos) {
+    return Todos.map((todo, index) => {
```
spirosikmd: Instead of using the `Todo` directly you would have set them in the state of the `App`. For example:

```js
this.state = {
  todos: Todos
};
```

todo-app/src/TodoStatus.js
```diff
@@ -0,0 +1,35 @@
...
+    let doneState = 'Not Done'
+    if (this.state.done) {
+      doneState = 'Done'
+    }
+
+    return (
+      <div>
+        <input type="checkbox"
+          checked={this.state.done}
+          onChange={this.handleChange}>
```
spirosikmd: Nice! You used `checked` and `onChange` 🎉

todo-app/src/TodoStatus.js
```diff
@@ -0,0 +1,35 @@
...
+  handleChange(event) {
+    this.setState({
+      done: !this.state.done
+    })
+  }
+
+
+  render() {
+    let doneState = 'Not Done'
+    if (this.state.done) {
```
spirosikmd: I like this rendering logic here, but you could have extracted it to a method, for example:

```js
renderDoneLabel(done) {
  return done ? 'Done' : 'Not Done';
}
```

and call it in the `label`.

todo-app/src/TodoStatus.js
```diff
@@ -0,0 +1,35 @@
...
+    super(props);
+    this.state = {
+      done: this.props.done || false,
+    }
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  handleChange(event) {
+    this.setState({
+      done: !this.state.done
```
spirosikmd: Updating the state with `setState` is the way to go, nice 👍 You could have used the `checked` property of the event target, `event.target.checked`. So you don't depend on the previous value of the state. You might run in problems with this, as the state [updates asynchronously](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous).

todo-app/src/App.js
```diff
@@ -1,13 +1,21 @@
...
+      return (
+        <TodoList key={index} text={todo.text} />
```
spirosikmd: Here you should have used a component that renders the text and checkbox in an `li` element. Now you are rendering for each todo item, one `ul` and one `li`, which results in the following

screen shot 2017-10-28 at 11 16 36

which is not the required result!

PR #10, opened 2017-10-20
Halim homework

HalimSD:


todo-app/src/App.js
```diff
@@ -1,13 +1,24 @@
...
+import React from 'react';
+import TodoList from './components/TodoList';
+import Todos from './components/todos';
 
-class App extends Component {
+export default class App extends React.Component {
+  renderTodos(Todos) {
+    return Todos.map((todo, index) => {
+      return (
+        <TodoList
```
spirosikmd: A better naming for this component as it accepts a single todo object is `Todo`? What do you think?

todo-app/src/components/TodoItem.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+export default class TodoItem extends React.Component {
+  render() {
+    return (
+      <div><ul>
```
spirosikmd: If each `TodoItem` renders a `ul` then for three items we will have three `ul` elements, which is not very nice, right? Ideally we want to have one `ul` and each item should render an `li` element?

todo-app/src/components/TodoList.js
```diff
@@ -0,0 +1,19 @@
...
+import TodoStatus from './TodoStatus';
+import TodoNumber from './TodoNumber';
+
+
+
+export default class TodoList extends Component {
+    
+    render() {
+        const { id, text, status } = this.props.todo;     
+        return (<div>
```
spirosikmd: Code formatting here is a bit off, makes it harder to read. Especially when we need to write more code and collaborate with other developers.

todo-app/src/todoItems.js
```diff
@@ -0,0 +1,19 @@
+import React, { Component } from 'react';
+import TodoDescription from './todoDescription'
+import TodoState from './todoState'
+
+class todo extends Component {
+    render() {
+        return (
+            <div className='todoList' className='todoState' className='Description'>
+                <ul>
```
spirosikmd: You haven't fixed the problem that for each todo item, we now render both `ul` and `li` elements.

todo-app/src/todoList.js
```diff
@@ -0,0 +1,15 @@
+let todosList = [
```
spirosikmd: You can remove this `todo-app/src/todoList.js` file as it is not used anymore, right?

todo-app/src/todoItems.js
```diff
@@ -0,0 +1,19 @@
+import React, { Component } from 'react';
+import TodoDescription from './todoDescription'
+import TodoState from './todoState'
+
+class todo extends Component {
```
spirosikmd: In this project we should follow the convention that component class names start with capital and the name is exactly the same as the file name. So in this case, the class name should `TodoItem` and the filename should be `TodoItem.js`.

You should do that for all your components!

todo-app/src/todoState.js
```diff
@@ -0,0 +1,37 @@
...
+        super(props);
+        this.state = {
+            checked: this.props.done
+        }
+        this.handleChange = this.handleChange.bind(this);
+    }
+
+    handleChange(event) {
+        let changed = !this.state.checked
+        this.setState({
```
spirosikmd: Great that you used `setState` here. But you should either use `event.target.value` to be sure you are using the most updated checked value. Or use the previous state value, for example:

```js
this.setState(prevState => ({
  checked: !prevState.checked
}));
```

to avoid any issues with [asynchronous state updates](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous).

todo-app/src/todoState.js
```diff
@@ -0,0 +1,37 @@
...
+            checked: changed
+        })
+    }
+
+    todoStatus() {
```
spirosikmd: Nice! Conditional rendering logic for the text!

PR #11, opened 2017-10-20
Amir react HW

ctsctz:


todo-app/src/App.js
```diff
@@ -1,13 +1,25 @@
...
 
-class App extends Component {
+export default class App extends React.Component {
   render() {
     return (
       <div>
-        Hello World!
+        <h1>ToDo List:</h1>
+        <div>
+          {action.map((action, index) => {
```
spirosikmd: Why have you used the term action? Naming things is very important in software development, so for consistency we should probably use something related to todo. What do you think?

todo-app/src/App.js
```diff
@@ -1,13 +1,25 @@
...
+            return (
+              <Tasks
```
spirosikmd: This should probably refer to a single item as it accepts a single action object. For example `Task`?

todo-app/src/Tasks.js
```diff
@@ -0,0 +1,20 @@
...
+import Done from './Done-action';
+
+
+export default class Tasks extends React.Component {
+    render() {
+        const { text, done } = this.props.task;
+        return (
+            <div>
+                <div>
+                    <info text={text} />
```
spirosikmd: React guidelines specify that custom components as in this case, should be named with [first capital letter](https://reactjs.org/docs/components-and-props.html#rendering-a-component). So `Info` instead of `info`.

PR #12, opened 2017-10-20
react first HW

aymansalah1:


todo-app/src/App.js
```diff
@@ -1,13 +1,22 @@
...
+import React from 'react'
+import ToDo from './ToDo.js'
+import ToDos from './toDos-DB.js'
 
-class App extends Component {
+export default class App extends React.Component {
+  renderToDos(toDos) {
+    return toDos.map((toDo, index) => {
+      return (
+        <ToDo key={index} toDo={toDo} />
```
spirosikmd: Nice that you used single case here for a `ToDo`!

todo-app/src/ToDo.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import TodoDesc from './TodoDesc.js'
+import ToDoStatus from './ToDoStatus.js'
+import './todo.css'
+
+
+export default class TODO extends React.Component {
```
spirosikmd: It would be nice for the class name to reflect the file name, for example `class ToDo`

todo-app/src/TodoDesc.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+import './todo.css'
+
+export default class TodoDesc extends React.Component {
+
+    render() {
+        return (
+            <dev className="toDo-Description" >* {this.props.todoDesc}</dev>
```
spirosikmd: What is `dev`? Probably a typo!

todo-app/src/ToDo.js
```diff
@@ -0,0 +1,24 @@
...
+    render() {
+        const { todoDesc, toDoStatus } = this.props.toDo
+        return (
+            <div className="toDoItem">
+                <div>
```
spirosikmd: You didn't really need the extra `div`s around the components. This should work as well

```jsx
<div className="toDoItem">
  <TodoDesc todoDesc={todoDesc} />
  <ToDoStatus toDoStatus={toDoStatus} />
</div>
```

todo-app/src/toDos-DB.js
```diff
@@ -0,0 +1,18 @@
+export let ToDos=[
+    {
+        todoDesc: "Get Out of Bed",
+        toDoStatus: "Done"
```
spirosikmd: You could have used a `boolean` property here for `toDoStatus`, could be either `true` or `false`.

todo-app/src/TodoDesc.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+import './todo.css'
+
+export default class TodoDesc extends React.Component {
+    render() {
+        return (
+            <dev className="toDo-Description" >* {this.props.todoDesc}</dev>
```
spirosikmd: Instead of using `*`, you could have used a `ul` element for the list and each todo item could be `li` element.

todo-app/src/App.js
```diff
@@ -1,13 +1,22 @@
-import React, { Component } from 'react';
+import React from 'react'
+import ToDo from './ToDo.js'
+import {ToDos} from './toDos-DB.js'
 
-class App extends Component {
+export default class App extends React.Component {
+  renderToDos(toDos) {
+    return toDos.map((toDo, index) => {
```
spirosikmd: Instead of rendering the `toDos` directly from the import, you should have set them in `App` state and use them from there. For example:

```js
export default class App extends React.Component {
  constructor(props) {
    super(props);
    this.state = {
      todos: ToDos
    };
  }

  // we don't need to modify renderToDos

  render() {
    return (
      <div>
        <h1>ToDo List</h1>
        {this.renderToDos(this.state.todos)}
      </div>
    )
  }
}
```

The will help us in updating and re-rendering. When for example the status of a todo change and we update the state, then React will now that it has to render the components that are affected by this change. You now have to do some tricks to accomplish this, which should be avoided (see `ToDoStatus ` comments).

todo-app/src/toDos-DB.js
```diff
@@ -0,0 +1,18 @@
...
+        todoDesc: "Brush Teeth",
+        toDoStatus: "Not Done"
+    },
+    {
+        todoDesc: "Eat Breakfast",
+        toDoStatus: "Not Done"
+    }
+]
+
+export default function updateStatus(index, status) {
```
spirosikmd: You shouldn't use this function here that updates the database directly. The fact that this is a JavaScript file is convenient, but what if this was a JSON file. You wouldn't be able to update the todos. There are other ways to accomplish this in React.

todo-app/src/ToDoStatus.js
```diff
@@ -0,0 +1,45 @@
...
+    componentDidMount() {
+        this.refs.chxToDoStatus.checked = (this.status.toDoStatus === "Done") ? true : false;
+    }
+    render() {
+        return (
+            <div className="toDo-Status-chx">
+                <div>
+                    <input
+                        type="checkbox"
+                        ref="chxToDoStatus"
```
spirosikmd: We should avoid using `ref`s in React. The way to accomplish this is by using a combination of `onChange` and `value` on the input, assuming that `status` is a boolean value. 

```js
export let ToDos = [
    {
        todoDesc: "Get Out of Bed",
        toDoStatus: true
    },
    {
        todoDesc: "Brush Teeth",
        toDoStatus: false
    },
    {
        todoDesc: "Eat Breakfast",
        toDoStatus: false
    }
];
```

For example:

```jsx
// Somewhere in the constructor 
this.state = {
  toDoStatus: this.props.toDoStatus
};

// Somewhere in render
<input
  type="checkbox"
  checked={this.state.toDoStatus}
  onChange={this.toggleCheckbox}
/>
```

In this case `toggleCheckbox` will be called with an `event` that contains the target with the new `checked` value, `event.target.checked`. Now that we have this value in the `toggleCheckbox` function, how can we modify the state without using `updateStatus` (basically we should remove this function 😄)??

todo-app/src/toDos-DB.js
aymansalah1: hi Spyros ,
 thanks for your feed back, could you please tell me a sample how to do the Update with Ract

spirosikmd: Hi @aymansalah1! Take a look at my comment at `todo-app/src/ToDoStatus.js`. You can use a combination of `state`, and input's `checked` and `onChange` attributes to accomplish this.

todo-app/src/App.js
```diff
@@ -1,21 +1,132 @@
...
+        todoDesc: ""
+      }
+    }
+    this.handleInputChange = this.handleInputChange.bind(this);
+    this.handleNewToDo = this.handleNewToDo.bind(this);
+    this.sendForm = this.sendForm.bind(this);
+    this.deleteToDo = this.deleteToDo.bind(this);
+    this.updateToDo = this.updateToDo.bind(this);
+    this.editToDo = this.editToDo.bind(this);
+  }
```
spirosikmd: Would be nice to have a new line after each function declaration, just to make it easier to read.

todo-app/src/App.js
```diff
@@ -1,21 +1,132 @@
...
+      }
+      return toDo;
+    });
+    this.setState({
+      ToDos: updatedToDos
+    })
+  }
+  deleteToDo(toDoIndex) {
+    let updatedToDos = this.state.ToDos
+    updatedToDos.splice(toDoIndex, 1)
```
spirosikmd: Here you are mutating the original array. Maybe you can use `filter` instead?

todo-app/src/App.js
```diff
@@ -1,21 +1,132 @@
...
+    this.setState({
+      ToDos: updatedToDos
+    })
+  }
+  renderToDos() {
+    let toDoItems = this.state.ToDos;
+    if (toDoItems.length > 0) {
```
spirosikmd: You can avoid the `else` statement if you reverse this condition `toDoItems.length === 0`.

todo-app/src/TodoDesc.js
```diff
@@ -2,9 +2,31 @@ import React from 'react'
 import './todo.css'
 
 export default class TodoDesc extends React.Component {
+    constructor(props) {
+        super(props)
+    }
+    editTodo = (event) => {
+        this.props.controlActions.editToDo(this.props.index)
```
spirosikmd: You could use the todo id instead of index. We are safer using something more unique instead of the index.

todo-app/src/TodoDesc.js
```diff
@@ -2,9 +2,31 @@ import React from 'react'
...
+        this.props.controlActions.deleteToDo(this.props.index)
+    }
+    handleInputChange(event) {
+
+    }
     render() {
         return (
-            <dev className="toDo-Description" >* {this.props.todoDesc}</dev>
+            <div className="toDo-Description">
+                <dev  >
```
spirosikmd: This should probably be a `div` 😄

todo-app/src/TodoDesc.js
```diff
@@ -0,0 +1,31 @@
...
+    render = () => {
+        let readOnlyOPT = {
+            'readOnly': this.props.isUpdating ? "" : "readOnly"
+        }
+        return (
+            <div className="toDo-Description">
+                <dev  >
+                    <input type="text" name="username" value={this.state.updatedToDoDesc} onChange={this.handleInputChange} {...readOnlyOPT} />
+                </dev>
+                <div><a href="#" className="editToDo" name="Edit" onClick={this.props.controlActions.editToDo} >{this.props.isUpdating ? "Cancel" : "Edit"}</a></div>
```
spirosikmd: We should not use `href="#"` for `a` tags. We als get a nice warning from React.

screen shot 2017-11-09 at 08 22 17

todo-app/src/App.js
```diff
@@ -1,13 +1,110 @@
...
+    this.setState({
+      ToDos: updatedToDos,
+      newToDo: {
+        todoDesc: ""
+      }
+    })
+  }
+  editToDo = (toDoIndex) => {
+    this.setState({
+      upatingTodoIndex:this.state.upatingTodoIndex?null:toDoIndex+1
```
spirosikmd: This logic is a bit tricky... When we write code, we write so we can understand it in the future and our colleagues can understand it as well! you are trying to do everything at once in one line. 

How would you explain this line of code to someone? How can you improve it? Maybe cancel action of a todo should trigger another action?

todo-app/src/App.js
```diff
@@ -1,13 +1,110 @@
...
+      return toDo;
+    });
+    this.setState({
+      ToDos: updatedToDos,
+      upatingTodoIndex: null
+    })
+  }
+  deleteToDo = (toDoIndex) => {
+    let updatedToDos = this.state.ToDos
+    updatedToDos.splice(toDoIndex, 1)
```
spirosikmd: We should not mutate the todos array here. How can you avoid that?

PR #13, opened 2017-10-21
myReact

mrbarri:


todo-app/src/App.js
```diff
@@ -1,13 +1,40 @@
...
+//       </div>
+//     );
+//   }
+// }
+
+// export default App;
+
+import React from 'react';
+import Tasks from './Task'
+import task from './taskDB'
```
spirosikmd: This could be named plural `tasks` to avoid confusion. Also it actually represents and array of tasks.

todo-app/src/App.js
```diff
@@ -1,13 +1,40 @@
...
+export default class App extends React.Component {
   render() {
     return (
       <div>
-        Hello World!
+        <h1>ToDo:</h1>
+        <div>
+          {task.map((task, index) => {
+            return (
+              <Tasks
```
spirosikmd: If this renders a single task then we should probably name it `Task` right? Also it would help a lot to keep naming consistent, for example instead of `Task` we name it `Todo` or `TodoItem`?

mrbarri: Thanks for feedback . I need to think first before name it later .

PR #15, opened 2017-10-21
Homework-Week2

Ramicode:


todo-app/src/App.js
```diff
@@ -1,13 +1,23 @@
...
   render() {
     return (
-      <div>
-        Hello World!
+      <div className='main'>
+        <h1>Todo List App</h1>
+        {TodosList.map((todo, index) => {
+          return (
+          <Todo
+          todoDescription={todo.todoDescription}
```
spirosikmd: You could pass the `todo` as a whole so you don't need to extend the props if `todo` gets more properties that we want to render inside `Todo`. So something like

```jsx
<Todo todo={todo} />
```

todo-app/src/todo.js
```diff
@@ -0,0 +1,20 @@
...
+import TodoDescription from './todoDescription'
+import TodoState from './todoState'
+import './App.css'
+
+class todo extends React.Component { 
+  render() {
+    return (
+      <div>
+        <div className='Description'>
+          <li><TodoDescription todoDescription={this.props.todoDescription}/></li>
```
spirosikmd: Why are you using `li` elements if they are not children of a `ul` element?

todo-app/src/todoState.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+
+class todoState extends React.Component {
```
spirosikmd: Component names should be named using [pascal case](http://wiki.c2.com/?PascalCase)!

todo-app/src/todoDescription.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+
+class todoDescription extends React.Component {
```
spirosikmd: Component names should be named using [pascal case](http://wiki.c2.com/?PascalCase)! And same the file name!

todo-app/src/App.js
```diff
@@ -1,13 +1,23 @@
...
+import './App.css'
 
 class App extends Component {
   render() {
     return (
-      <div>
-        Hello World!
+      <div className='App'>
+        <h1>Todo List App</h1>
+        {todosTable.map((todo, index) => {
```
spirosikmd: Instead of using the `todosTable` directly you could have set them in the state of the `App`. For example:

```js
this.state = {
  todosTable: todosTable
};
```

todo-app/src/todoState.js
```diff
@@ -0,0 +1,45 @@
...
+    if (this.state.checked) {doneOrNot = 'Done'}
+    return doneOrNot
+  }
+
+  render() {
+    return (
+      <div>
+        <input type="checkbox"
+          checked={this.state.checked}
+          onChange={this.handleChange}>
```
spirosikmd: Nice! You used `checked` and `onChange` 🎉

todo-app/src/todosTable.js
```diff
@@ -0,0 +1,15 @@
+let todosTable = [
+    {
+        todoDescription: 'Get out of bed',
+        done: true
```
spirosikmd: `done` as `boolean` value, nice 👍

todo-app/src/todoState.js
```diff
@@ -0,0 +1,45 @@
...
+        </input>
+        <label style={this.changeStyle()}>
```
spirosikmd: Nice! You used two extra functions with rendering logic. This is the way to go 👍

todo-app/src/todoState.js
```diff
@@ -0,0 +1,45 @@
...
+    super(props);
+    this.state = {
+      checked: this.props.done
+    }
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  handleChange(event) {
+    let changed = !this.state.checked
+    this.setState({
```
spirosikmd: Updating the state with `setState` is the way to go, nice 👍 You could have used the `checked` property of the event target, `event.target.checked`. So you don't depend on the previous value of the state. You might run in problems with this, as the state [updates asynchronously](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous).

todo-app/src/App.js
```diff
@@ -1,13 +1,152 @@
...
+    this.changeButtonsVisibility = this.changeButtonsVisibility.bind(this)
+  }
+
+  componentDidMount() {
+    setTimeout(() => {
+      this.setState({ todos })
+    }, 1000)
+  }
+
+  changeTodoStage(todoStage, todoID) {
```
spirosikmd: Nice that you used this function! This is quite similar with `updateTodo`. The only thing that changes is `{ stage: todoStage }`. Maybe you can generalize this one and end up with one function?

todo-app/src/App.js
```diff
@@ -1,13 +1,152 @@
...
+  // toggle the Visibility of cancel and update buttons
+  changeButtonsVisibility(show, clickedBtn) {
+    this.setState({
+      visibility: show,
+      // save the clicked edit button to render cancel and update buttons exclusively
+      clicked: clickedBtn
+      ///////////////////////////////////////////////////////////////////////////////
+    })
+  }
+//////////////////////////////////////////////////////
```
spirosikmd: What are these long lines of `////`?

todo-app/src/App.js
```diff
@@ -1,13 +1,152 @@
...
+    })
+    const updatedTodos = [...this.state.todos, updatedNewTodo]
+    this.setState({
+      todos: updatedTodos
+
+    })
+  }
+
+  deleteTodo(deletedTodoID) {
+    const newTodos = this.state.todos.filter(todo => {
```
spirosikmd: Great that you used `filter`!

todo-app/src/App.js
```diff
@@ -1,13 +1,152 @@
...
+  onFormSubmit(event) {
+    event.preventDefault()
+    const updatedNewTodo = Object.assign({}, this.state.newTodo, {
+      id: this.state.todos.length + 1,
+      done: false
+    })
+    const updatedTodos = [...this.state.todos, updatedNewTodo]
+    this.setState({
+      todos: updatedTodos
+
```
spirosikmd: You could set the `newTodo.description` here to empty string after successfully adding a todo.

todo-app/src/App.js
Ramicode: to mark the end of change :) ... but as of your response I might have not to use them again :)

todo-app/src/App.js
Ramicode: now that I see it again I see what you mean

PR #18, opened 2017-11-02
myReact2

mrbarri:


todo-app/src/todo.js
```diff
@@ -0,0 +1,20 @@
+import React, { Component } from 'react';
+import Description from './description'
+import TodoBox from './todoBox'
+import './App.css'
+
+class todo extends Component {
+    render() {
+        return (
+            <div class = 'todoList'>
```
spirosikmd: In React we should use `className` instead of `class` due to [jsx](https://reactjs.org/docs/introducing-jsx.html#specifying-attributes-with-jsx).

todo-app/src/todo.js
```diff
@@ -0,0 +1,20 @@
...
+                <div class = 'todoBox'>
+                    <li><TodoBox done={this.props.done}/></li>
+                    </div>
+                    <div class ='Description'>
+                        <li>Description description={this.props.description} /></li>
```
spirosikmd: Why did you add this `/>` as part of the text?

todo-app/src/todoBox.js
```diff
@@ -0,0 +1,45 @@
...
+    let doneOrNot = 'Not done'
+    if (this.state.clicked) {doneOrNot = "Done"}
+    return doneOrNot
+}
+
+render() {
+    return (
+        <div>
+        <input type = "checkbox"
+        clicked = {this.state.clicked}
```
spirosikmd: There is no `clicked` attribute for [`input checkbox`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox) element. Maybe you wanted to use something else?

todo-app/src/todoBox.js
```diff
@@ -0,0 +1,45 @@
...
+        onClick = {this.handleChange}>
```
spirosikmd: We should use another event handler here and not `onClick` to respond to change. Which one is it?

todo-app/src/todoBox.js
mrbarri: I've checked my code 2 3 times but I think I did mistake I did put some code while we were using last Sunday, reading all code is hard

todo-app/src/todo.js
mrbarri: @spirosikmd Thanks for feedback . It works here and completely fine in the browser . I don't know where is the problem . so I will read the code again

PR #21, opened 2017-12-12
Eyad-homework1

meyadd:


isaachinman:
Hi @meyadd, thanks for being so fast with this!

Overall, it looks good. Your `todo-app` functions as it is supposed to.

A few points to fix:

1. You will notice if you open the browser console that React is warning you about keys. This is because you're iterating over your `todo_arr` with `map`, but not giving your returned top-level JSX keys. You can check out [this section](https://reactjs.org/docs/lists-and-keys.html) of the React docs for more info. You need to give your `Task` components a unique key so React can tell when they are changed or removed.

2. Some minor formatting corrections would be nice, mostly in `App.js` and `task.css` JSX is customarily indented like HTML would be, and with CSS we usually add spaces after the colon to make it a little easier to read.

Otherwise good work, you made it look easy like you said!

todo-app/src/App.js
```diff
@@ -1,13 +1,33 @@
...
 
 class App extends Component {
+	
   render() {
     return (
       <div>
-        Hello World!
+        <h1>Todo list</h1>
+		<ul className='task-list'>
+			{todo_arr.map(item => <Task task={item} />)}
```
isaachinman: You will notice if you open the browser console that React is warning you about keys. This is because you're iterating over your `todo_arr` with `map`, but not giving your returned top-level JSX keys.

You can check out [this section](https://reactjs.org/docs/lists-and-keys.html) of the React docs for more info. You need to give your `Task` components a unique key so React can tell when they are changed or removed.

todo-app/src/task.css
```diff
@@ -0,0 +1,17 @@
+.task-list{
```
isaachinman: Some minor formatting corrections would be nice, mostly in `App.js` and here in `task.css` JSX is customarily indented like HTML would be, and with CSS we usually add spaces after the colon to make it a little easier to read.

PR #22, opened 2017-12-14
Homework week1

Elias011:


todo-app/src/App.js
```diff
@@ -1,13 +1,21 @@
...
-class App extends Component {
-  render() {
-    return (
-      <div>
-        Hello World!
-      </div>
-    );
-  }
+export class Todo extends React.Component{
+    Constructor(){
```
isaachinman: I guess you must have been looking at React component examples somewhere? For this homework, because we don't have any component state or functions, we don't need a constructor.

We will get more into functions and state this coming week. Just know that a class `constructor` is _not necessary_ for a React component, but if you do choose to use it, it's lowercase: `constructor`.

todo-app/src/App.js
```diff
@@ -1,13 +1,21 @@
...
+        
+    }
+    
+    render(){
+        return (
+            <div className = "todoApp">
+            <h1>Todo App</h1>
+            <AppDate tasks={this.props.tasks}/>
+            <AppTask tasks={this.props.tasks}/>
```
isaachinman: What was your reasoning for splitting the tasks and dates into two separate children like this? For now they render on top of each other and are "together", but what if we had a very long list of 100 items? It's fine to split date and task content into two separate components - say we call them `TodoTask` and `TodoDate` - but they should then be wrapped in a higher-order component just called `Todo`, and you can loop through the `Todo` components instead.

todo-app/src/App.js
```diff
@@ -1,13 +1,21 @@
-import React, { Component } from 'react';
+import React from 'react';
+import {AppTask} from './AppTask';
```
isaachinman: When creating React components in a single file, it's common practice to set that component as the default export, so you can just do:

    import AppTask from './AppTask'

The only time where it's appropriate to use named exports like this is if we were exporting _multiple_ components from a single file.

Check out [this Stackoverflow post](https://stackoverflow.com/questions/33611812/javascript-es6-export-const-vs-export-default), it's a pretty good explanation.

In your case, all you have to do is add `default` to your exported components (I'll also leave a comment there), and change the import syntax here.

todo-app/src/AppDate.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+export class AppDate extends React.Component{
```
isaachinman: This is where you'd add a `default` keyword.

todo-app/src/AppTask.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+export class AppTask extends React.Component{
+    render(){
+       let date = this.props.tasks.map((elem, i)=>{
+           return <div>
```
isaachinman: When you run your app via `yarn start`, try opening the browser console. You'll see two warnings from React. Both of them have to do with creating children in arrays (your `map` functions).

When creating React components in a loop, it's extremely important to give the top-level JSX tag a `key` prop. This `key` prop's value must be unique to that element.

While using index isn't great practice, we can leave it for the time being.

Your main problem here is that you provided the `key` prop to a child _inside_ your returned JSX, so React has no idea it's there. It needs to be a prop of the _highest level_ JSX returned, so in this case on the `div`.

todo-app/src/AppDate.js
```diff
@@ -0,0 +1,17 @@
...
+    render(){
+    let items = this.props.tasks.map((elem, i)=>{
+        return <div>
```
isaachinman: Here's the other case of the `key` prop being nested and not on the top-level JSX.

todo-app/src/App.js
Elias011: hahaha yes you are right it was a tutorial on youtube!!

Elias011:
Thanks for the advices Isaac.
I will work on them

Elias011:
Ok that makes sense

PR #23, opened 2017-12-15
Week1 version1

pesh12:
tried to implement, how we did in the class.


pesh12:
Isaac, I have closed the other PR.
Actually the other one was very basic.

isaachinman:
@pesh12 Thanks! By the way, make sure to tag someone by using the `@` symbol when you mention their name so that I get an email.

pesh12:
@isaachinman 
Actually I tried but, your name didn't appear after @ 
But now appeared. Anyways :))
Thanks

todo-app/public/index.css
```diff
@@ -0,0 +1,31 @@
+h2{
```
isaachinman: A few formatting issues in your CSS here. It's OK to indent with 2 spaces or 4 spaces, but choose one and stick to it. Also, add spaces after your class names and before the curly brackets.

todo-app/public/index.html
```diff
@@ -5,6 +5,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
     <meta name="theme-color" content="#000000">
     <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico">
+    <link rel="stylesheet" href="index.css">
```
isaachinman: This is not the right way to include styles in your `create-react-app` project!

You can read [the docs here](https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#adding-a-stylesheet) about how styles should work within `create-react-app` projects.

We can also discuss this in person, but basically there is a relatively complicated build toolchain set up to allow you to make your styles modular, and to be able to import them into your js files in a logical way. The days of having a single massive `index.css` are over, we use modular styles now.

A great example is if you build a component, and then suddenly need to use this component in another project. We can simply copy/paste the entire component directory:

    |- MyComponent
      ---- MyComponent.js
      ---- MyComponent.styles.js

And not have to worry about messy global styles.

todo-app/src/TodoApp.js
```diff
@@ -0,0 +1,15 @@
+import React, { Component } from 'react';
+import TodoList from './TodoList'
```
isaachinman: Another picky comment: missing a semicolon here (and a few other places). Since we're human, it's very easy to miss things like this. That's where linters come in handy.

PR #24, opened 2017-12-15
Week1 version2

pesh12:
simpler version of todo-app


isaachinman:
Hey @pesh12, I see you've got two open PRs. Which one would you like me to take a look at? Please just close whichever PR isn't up for consideration.

PR #25, opened 2017-12-15
Week1 hw

itemah26:


todo-app/src/Task.js
```diff
@@ -0,0 +1,17 @@
...
+import TaskText from './TaskText';
+import TaskDeadLine from './TaskDeadLine';
+
+export default class Task extends React.Component {
+
+  render() {
+
+    return (
+      <li className="task-list"> 
+        <TaskText text={this.props.task.text}/>
```
isaachinman: Would be nice to see some object destructuring here to prevent repetition of `this.props.task`.

todo-app/src/TaskDeadLine.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+
+export default class TaskDeadLine extends React.Component {
+     render() {
+          return (
+               <div className="task-deadline">
+               {this.props.deadline}
```
isaachinman: It's customary to indent JSX just like you would indent HTML.

PR #26, opened 2017-12-16
Added homework for week 1

aminmahrami:


todo-app/src/TodoList.js
```diff
@@ -0,0 +1,48 @@
...
+ },
+ {
+  id: '3',
+  task: "Eat breakfast",
+  time: "Fri Sep 15 2017" 
+ }
+]
+
+export default class TodoList extends React.Component {
+    render() {
```
isaachinman: It is customary to have the `render` function be the _last_ function in a React component. Of course this is 100% a matter of preference, but it's how most devs do it. You can check out [this link](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md) for examples within an `eslint` rule.

todo-app/src/TodoApp.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import TodoList from './TodoList';
+
+export default class TodoApp extends React.Component {
+
+    render() {
+        return(
```
isaachinman: There are some inconsistencies with your indentation. It's OK to use either 2 spaces or 4 spaces for indentation, but choose one and stick to it!

PR #27, opened 2017-12-17
Ji homework week1

JiDarwish:
This is the homework for week 1, simple todo-app with React


todo-app/public/index.html
```diff
@@ -4,6 +4,7 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
     <meta name="theme-color" content="#000000">
+    <link rel="stylesheet" href="style.css">
```
isaachinman: This is not the right way to include styles in your `create-react-app` project!

You can read [the docs here](https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#adding-a-stylesheet) about how styles should work within `create-react-app` projects.

We can also discuss this in person, but basically there is a relatively complicated build toolchain set up to allow you to make your styles modular, and to be able to import them into your js files in a logical way. The days of having a single massive `index.css` are over, we use modular styles now.

A great example is if you build a component, and then suddenly need to use this component in another project. We can simply copy/paste the entire component directory:

    |- MyComponent
      ---- MyComponent.js
      ---- MyComponent.css

And not have to worry about messy global styles.

todo-app/public/style.css
```diff
@@ -0,0 +1,49 @@
...
+body {
+    background: aliceblue;
+    text-align: center;
+}
+
+ul {
+    list-style: none;
+}
+
+.todosWrapper {
```
isaachinman: You're using both camelcase and dashes for your CSS class names here. Would be better to pick one naming convention and stick to it.

todo-app/src/components/Todos/List-todos.js
```diff
@@ -0,0 +1,69 @@
...
+
+import React from 'react';
+import TodoItem from './Todo-item';
+
+
+let todosInfo = [
+    {
+        id: 1,
+        author: 'person 1',
+        date: new Date().toLocaleString(),
```
isaachinman: It would be better to just set our date as `new Date` here, and use `toLocaleString` when we want to display it in some UI. Maintaining the full date object as opposed to immediately converting it to a string allows us much more flexibility.

todo-app/src/components/Todos/List-todos.js
```diff
@@ -0,0 +1,69 @@
+// jshint esnext: true
+
+import React from 'react';
+import TodoItem from './Todo-item';
```
isaachinman: It would be a little nicer to name your component files exactly as the component is named, so that it's immediately obvious what it is. So something like:

    import TodoItem from './TodoItem';

todo-app/src/components/Todos/Todo-Avatar.js
```diff
@@ -0,0 +1,11 @@
+// jshint esnext: true
+
+import React from 'react';
+
+export default class Avatar extends React.Component {
+    render(){
+        return (
+            <img src={this.props.avatar}/>
```
isaachinman: If you run your app with `yarn start`, you'll see this warning in the console:

    ./src/components/Todos/Todo-Avatar.js
      Line 8:  img elements must have an alt prop, either with meaningful text, or an empty string for decorative images  jsx-a11y/alt-text

This is an accessibility thing, you can check out [docs here](https://www.w3schools.com/tags/att_img_alt.asp).

todo-app/public/index.html
JiDarwish: Okay I get it, it seems tempting to have a small amount of css in each file instead of a huge one, but I wonder if it will become hard if there was like a container component and a contained component and they had some relation in the styling like Flex-box or something.

todo-app/public/style.css
JiDarwish: Yeah you're right.

todo-app/src/components/Todos/List-todos.js
JiDarwish: The same thing with what to use camel case or something else, I think I'll just do camel case from now on it's usable in the js code as well

todo-app/public/index.html
isaachinman: Yeah, that's a fair point. Styling React components is the one thing where there is no real "standard" the community has decided upon. What you're bringing up depends upon your methodology for managing React components in a hierarchical way. [Here's one example](https://medium.com/@yejodido/atomic-components-managing-dynamic-react-components-using-atomic-design-part-1-5f07451f261f), although I don't follow it or know enough to recommend it.

Basically - it's up to you.

My point was that it's an anti-pattern to include your own `index.css` in the `index.html` directly (or really put any CSS in the `public` dir), webpack does this for you. As long as you use the build tools in CRA appropriately, the rest is up to you!

todo-app/src/components/Todos/List-todos.js
isaachinman: What do you mean? React components will always need an uppercase first letter. Besides React components, yeah, camel case is a safe bet for all JS.

todo-app/src/components/Todos/ListTodos.js
```diff
@@ -1,43 +1,47 @@
...
 
 
 let todosInfo = [
     {
         id: 1,
         author: 'person 1',
-        date: new Date().toLocaleString(),
+        date: new Date(),
         avatar: 'https://d30y9cdsu7xlg0.cloudfront.net/png/770852-200.png',
+        alt: this.author + ' image',
```
isaachinman: If your logic for the `alt` attribute is simply `todo.author + ' image'`, you can do that directly inside the component and save repetition here.

isaachinman:
Very nice work. I like your attention to detail!

PR #28, opened 2017-12-20
Ji week2's homework

JiDarwish:
Here's the homework for this week, I decided to make a new version of it.
@isaachinman if you had time feedback would be much appreciated.


todo-app/src/Components/todoItem.js
```diff
@@ -0,0 +1,21 @@
...
+
+
+export default class TodoItem extends React.Component {
+    render (){
+        const {description, deadline, done, id } = this.props.data;
+        return (
+            <li className="todoItem">
+                <div>
+                    <span>{deadline}</span>
+                    <label for={`item${id}`} className={done ? 'markDone' : 'markUndone'}></label>
```
isaachinman: If you open your browser console after running your app, you'll see this error:

    Warning: Invalid DOM property `for`. Did you mean `htmlFor`?

Check out the [React docs here](https://reactjs.org/docs/dom-elements.html#htmlfor). Because `for` is a reserved word, you have to use `htmlFor` instead with JSX.

todo-app/src/Components/todoItem.js
```diff
@@ -0,0 +1,21 @@
...
+                    <input type="checkbox" id={`item${id}`} onClick={() => this.props.handleIsItDone(this.props.data)}/>
```
isaachinman: It's a matter of personal preference, but if you're declaring `item${id}` more than once, you might declare it in your `render` function before the return, like:

    const inputID = `item${id}`

todo-app/src/Components/todoItem.js
```diff
@@ -0,0 +1,21 @@
...
+                </div>
+                <p className={done ? 'done' : 'notDone'}>{description}</p>
```
isaachinman: Nice! Conditional class names!

todo-app/src/Components/todoList.js
```diff
@@ -0,0 +1,57 @@
...
+const todos = todosData.todos;
+
+
+export default class TodoList extends React.Component {
+
+    state = {
+        todos,
+    }
+
+    handleIsItDone = (todoItem) => {
```
isaachinman: It's probably better to call this function something like `handleToggleDone` or something, so that it's more obvious what the function actually does. With the current name, it makes me think the function just _checks_ if the thing is done, not actually changes it.

todo-app/src/App.js
```diff
@@ -1,13 +1,18 @@
-import React, { Component } from 'react';
+//jshint esnext: true
 
-class App extends Component {
+import React from 'react';
+
+import Todos from './Components/todoList';
```
isaachinman: I think we had a miscommunication on last week's homework about naming conventions. React component files are normally named exactly as the component is named, so in this case it would be `TodoList`. The uppercase first letter tells us immediately that it's a component.

Also, if you are calling your component `TodoList` via naming the file, and also naming the component inside the file, you should call it the exact same when you import it, eg:

    import TodoList from './Components/TodoList';

todo-app/src/Components/styles/todoItem.css
```diff
@@ -0,0 +1,44 @@
...
+    flex-direction: row;
+}
+
+.todoItem > * {
+    text-align: center;
+    flex: 3;
+    padding: 1rem;
+}
+
+.todoItem>*:first-child {
```
isaachinman: Careful, this is not a great way to write CSS. Since the `first-child` you are targeting here is a `div` which you are returning in your own JSX, it would be better to just give that `div` a class name.

todo-app/src/Components/todoList.js
```diff
@@ -0,0 +1,57 @@
...
+                <TodoItem 
+                data={item}
+                handleIsItDone={this.handleIsItDone}
+                />
+            </React.Fragment>
+        )})
+
+    }
+
+
```
isaachinman: Watch out for lots of empty lines in a row. Most linters will complain about this.

todo-app/src/Components/todoList.js
```diff
@@ -0,0 +1,57 @@
...
+            <ul className="todosContainer">
+                {this.state.todos.length === 0 ? //is it empty?
+                    <h1>No items to show</h1>
+                    :
+                    this.renderAllTodos()}
+            </ul>
+        )
+    }
+    
+    renderAllTodos = () => {
```
isaachinman: Not sure if it was us discussing this last week, but typically devs will put the `render` function last inside React components. Again it's a matter of preference, but you can check out this [eslint rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md) to see some reasoning for this convention.

PR #29, opened 2017-12-21
Eyad-homework2

meyadd:


todo-app/src/App.js
```diff
@@ -1,10 +1,49 @@
 import React, { Component } from 'react';
+import Task from './Task';
+import tasks from './task.json';
 
 class App extends Component {
```
isaachinman: Not terribly important, but just so you know, you can do your exporting inline on a class:

    export default class App extends Component

I know you probably pulled some of these examples from a tutorial, but in other files (`TaskDate.js`) you're exporting inline and here you're not, so just choose one method and use it throughout.

todo-app/src/App.js
```diff
@@ -1,10 +1,49 @@
...
+  
+  handleDoneChange(doneValue, taskId) {
+    const updatedtasks = this.state.tasks_arr.map(task => {
+      if (taskId === task.id) {
+        return Object.assign({}, task, {done: doneValue});
+      }
+      return task;
+    });
+    this.setState({
+      tasks_arr: updatedtasks
```
isaachinman: Would be nice to see a uniform method for variable/property naming. Underscoring or camel case are both fine.

todo-app/src/Task.js
```diff
@@ -0,0 +1,35 @@
...
+    this.markAsDone = this.markAsDone.bind(this);
+  }
+
+  markAsDone(event) {
+    const checked = event.target.checked;
+    this.props.onDoneChange(checked, this.props.task.id);
+  }
+  
+  render() {
+    const { description ,deadline , done } = this.props.task;
```
isaachinman: Watch out for the commas/spaces here. I'd definitely recommend using a linter like eslint to automatically check formatting like this.

todo-app/src/Task.js
```diff
@@ -0,0 +1,35 @@
...
+export default class Task extends React.Component {
+	
+  constructor(props) {
+    super(props);
+
+    // Events and bind to `this`
+    this.markAsDone = this.markAsDone.bind(this);
+  }
+
+  markAsDone(event) {
```
isaachinman: This function should probably be renamed, as it does not _only_ mark things as done. It toggles their "done" state.

meyadd:
Thanks Isaac for your valuable feedback
I have updated the homework

PR #30, opened 2017-12-22
Week2 Project

pesh12:
This time, I took care of indentation :)


pesh12:
Thanks for comments.
That is easier to change only the state. Actually that is exactly what we did last Sunday 
I am gonna work on this asap.

pesh12:
The delete handler is still there, just done-undone eventhandler added.

todo-app/src/TodoItem.js
```diff
@@ -6,15 +6,23 @@ import './index.css';
 export default class TodoItem extends React.Component {
 
   onDelete = () => this.props.onDelete(this.props.todoItem.id)
+
+  onDone = () =>  this.props.onDone(this.props.todoItem.id)
```
isaachinman: What's happening here? You're redeclaring `onDone` inside your child component `TodoItem`, which is fine, but you never use `this.onDone` - you're still using the `onDone` function from the `props` with an inline anonymous function:

    onClick={() => onDone(todoItem.id)}

Both ways are totally fine, but choose one. In this case both `this.onDelete` and `this.onDone` are never used.

todo-app/src/TodoList.js
```diff
@@ -0,0 +1,52 @@
...
+          todo.done = !todo.done
+        }
+        return todo
+      })
+    })
+  } 
+ 
+  render() {
+    const { todos } = this.state
+    if(todos.length == 0){
```
isaachinman: Careful here. If you run your app and open the Chrome devtools, you'll see this warning:

    webpackHotDevClient.js:241 ./src/TodoList.js
    Line 33:  Expected '===' and instead saw '=='  eqeqeq

You should always be using "strict equality comparison" unless there's a good reason not to. Here's a [StackOverflow question](https://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons) about it, and there's a bunch of info on the topic.

PR #31, opened 2017-12-24
Week2 - Last edits

pesh12:
Commented out the extra declared functions in the component.
Just figured out both ways "decaring - then using  way" and "inline anonymous way" 
Thanks @isaachinman


isaachinman:
Nice, that's exactly right. Just one note: be careful with comments in JSX. I actually have no idea how your app is compiling without errors with those `//` line comments, I've never seen that before. Typically people don't really use comments in JSX, or if they do, it's with a `{/* comment */}` syntax.

PR #32, opened 2017-12-25
Homework week2

Elias011:
Sorry I am late because it is christmas time
Merry christmas

### DIFF
--- a/todo-app/src/App.js
+++ b/todo-app/src/App.js
@@ -1,39 +1,39 @@
 import React, { Component } from 'react';
 import Task from './Task' 
-
-
-const tasks = [
-     {
-          id: "1",
-          description: "Wake up",
-          deadline: "Wed Sep 13 2017"
-  },
-     {
-          id: "2",
-          description: "React HomeWorks",
-          deadline: "Today"
-  },
-     {
-          id: "3",
-          description: "Breakfast",
-          deadline: "Today before 12:00"
-  }
-];
-
-
+import todoItems from './todo.json';
 
 
 export default class App extends Component {
-	
+    state = {
+    todos: todoItems
+  }
+
+handleTaskDone = (todoID)=>{
+    this.setState({
+            todos: this.state.todos.map(todo =>{
+                if (todo.id === todoID){
+                    todo.done =! todo.done
+                }
+                return todo
+            })
+        })
+    }
   render() {
+      const{todos} = this.state
     return (
       <div>
         <h1>Todo App</h1>
 		<ul className='tasks'>
-			{tasks.map(task => <Task key={task.id} task={task}/>
-    )}
+         {todos.map((todoItem, i)=>(
+        <Task key = {todoItem.id} handleTaskDone=
+        {this.handleTaskDone} todoItem = {todoItem}/>
+    
+    ))}
 		</ul>
       </div>
     );
   }
+  
+  
+      
 }
--- a/todo-app/src/App.js
+++ b/todo-app/src/App.js
@@ -1,13 +1,21 @@
-import React, { Component } from 'react';
+import React from 'react';
+import {AppTask} from './AppTask';
+import {AppDate} from './AppDate';
 
-class App extends Component {
-  render() {
-    return (
-      <div>
-        Hello World!
-      </div>
-    );
-  }
+export class Todo extends React.Component{
+    Constructor(){
+        
+    }
+    
+    render(){
+        return (
+            <div className = "todoApp">
+            <h1>Todo App</h1>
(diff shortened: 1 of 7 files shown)


todo-app/src/App.js
```diff
@@ -1,13 +1,39 @@
...
+import todoItems from './todo.json';
 
-class App extends Component {
+
+export default class App extends Component {
+    state = {
+    todos: todoItems
+  }
+
+handleTaskDone = (todoID)=>{
```
isaachinman: The name `handleTaskDone` is not accurate, because it marks todos as done _and also_ as not done. A more accurate name would be something like `handleToggleTodoStatus`, or something.

Also, be careful with your formatting! This line should be indented, and there should be spaces around your `=>`.

todo-app/src/App.js
```diff
@@ -1,13 +1,39 @@
...
+    this.setState({
+            todos: this.state.todos.map(todo =>{
+                if (todo.id === todoID){
+                    todo.done =! todo.done
```
isaachinman: Not sure if this was a simple mistake or a misunderstanding. The bang should be alongside the value itself, not next to the equals sign. There is the case of `!=` for (loose) equality comparison, it's important to note that this is totally different.

todo-app/src/Task.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import AppTask from './AppTask';
```
isaachinman: What is your `AppTask` component? You are importing it here, but never use it. If it's not used anymore, it should be removed, or at least not imported.

todo-app/src/todo.json
```diff
@@ -0,0 +1,20 @@
+[
+ {
+          "id": "1",
+          "description": "Wake up",
+          "deadline": "Wed Sep 13 2017",
+          "done":"true"
```
isaachinman: Check out the [JSON data types](https://www.w3schools.com/js/js_json_datatypes.asp). You need to be using booleans here, not `"true"` as a string - this is getting into very dangerous territory for things like equality comparison etc.

todo-app/src/Task.js
```diff
@@ -0,0 +1,41 @@
...
+  render() {
+  const{todoItem, handleTaskDone} = this.props;
+       return (
+            <li className="task-list">
+             <span>{todoItem.done ? <input type = 'checkbox' onClick={()=>handleTaskDone(todoItem.id)}/>
+              :
+              <input type = 'checkbox' checked onClick={()=> handleTaskDone(todoItem.id)}/>   
+             }
+                 </span>
+            <span> {todoItem.done? <span className = 'task-done'>
```
isaachinman: You've got a few problems here.

First of all, your `todoItem.done` value is displaying backwards - you are showing an empty checkbox when `todoItem.done` is `true`, and a full checkbox when it's `false`.

Second, if you run your app and open the Chrome console and click a checkbox, you'll see:

    proxyConsole.js:56 Warning: Task is changing an uncontrolled input of type checkbox to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components

You can check out the link provided in that warning to see what this is all about. Basically, you are first providing an uncontrolled input (a checkbox with no `checked` attribute specified), and then replacing it with a controlled input (a checkbox with a `checked` attribute set to true, ie it can never be unchecked).

So, what's the root of the problem? You are using a ternary to split rendering based on the `todoItem.done` value, which controls whether or not a checkbox is checked. But this is unnecessary. The "React" way to do this is:

    <input type='checkbox' checked={todoItem.done} onClick={() => handleTaskDone(todoItem.id)} />

No ternary, just pass the checked value directly.

Elias011:
Correction of my homework

isaachinman:
You still have the problem of your `done` status being backwards. I've attached a screenshot to show what I mean.

screen shot 2017-12-27 at 14 56 55

Elias011:
Sorry Isaac but actually i didn't get it!
could you explain a little bit?

isaachinman:
Of course, no problem. Basically, your app is initialising with all three todos marked as `"done": true` (coming straight from your JSON), but the visual display of the "done" value is backwards. The checkboxes are all unchecked and the text is not crossed out. Only when I click the checkbox and set the `done` property to `false` does the checkbox become checked and the text is crossed out. Does that make sense? It's just backwards, should be an easy issue to fix.

Elias011:
Is it now better?

isaachinman:
Yup, looks good.

PR #33, opened 2018-01-07
Homework week3 Ji

JiDarwish:
Here's homework for this week.
This one was very tricky so same as always feedback is very welcome.


todo-app/src/Components/AddTodoItem.js
```diff
@@ -0,0 +1,99 @@
...
+import React, { Component } from 'react'
+
+import Button from './Button';
+import InputField from './InputField';
+import './styles/addTodoItem.css';
+
+export default class AddTodoItem extends Component {
+
+    state = {
+        adding: false,
```
isaachinman: It's a little picky, but I might change this to something like `active`, as the component seems to have two states: active and inactive.

todo-app/src/Components/AddTodoItem.js
```diff
@@ -0,0 +1,99 @@
...
+        toBeAddedDeadline: ''
+    }
+
+
+
+    toggleAddFrom = event => {
+        this.setState({ adding: !this.state.adding });
+
+        //clearing fields for if it's a cancel
+        this.setState({
```
isaachinman: Why do you have two `setState` calls here? You're not clearing the fields _only_ on a cancel, you're clearing them in both cases. So you should either combine your `setState` calls into a single call, or wrap the second call in a condition.

Also check the name of the function itself - small typo.

todo-app/src/Components/todoItem.js
```diff
@@ -0,0 +1,94 @@
...
+import InputField from './InputField';
+
+
+export default class TodoItem extends Component {
+
+    handleRemove = () => { 
+        this.props.handleRemove(this.props.data.id);
+    }
+
+    handleToggleEdit = () => {
```
isaachinman: In my opinion, it makes it very confusing to have both local and props-based functions which are named the same. But this is especially the case here, because of the argument signature of the original `handleToggleEditing` in your `App` component.

todo-app/src/App.js
```diff
@@ -1,13 +1,83 @@
...
+
+  
+  state= {
+    todos: [],
+    isEditing: null
+  }
+
+
+  handleToggleEditing = (id) => {
+    this.setState({ isEditing: id || null })
```
isaachinman: This is an interesting choice. It is a pattern that used to be pretty common (maybe it still is) - if you do not include an argument, the function becomes a "cancel" action.

This might work in some cases, but I think it's a bad idea here. Because this function is passed down through several children, it should be immediately clear what it does to limit confusion. I'd recommend either adding a second argument, like:

    handleToggleEditing = (id, cancelEditing: boolean) =>

Or just declare a second cancel function in a single line, like:

    handleCancelEditing = () => this.handleToggleEditing(null)

todo-app/src/App.js
```diff
@@ -1,13 +1,83 @@
...
+  }
+
+  handleAddTodo = (text, deadline) => {
+    
+    const newItem = {
+        id: this.state.todos.length + 1 || 1,
```
isaachinman: We very briefly mentioned this on Sunday. What if I create three todos from scratch, so that I have a situation like this:

    [
      { title: "Thing 1", id: 1 },
      { title: "Thing 2", id: 2 },
      { title: "Thing 3", id: 3 },
    ]

Now imagine that I delete "Thing 2", and try to add a new todo. What will happen? You can start up your own app here and try it, you'll see the error in the dev console.

todo-app/src/App.js
```diff
@@ -1,13 +1,83 @@
...
+      if (item.id === id){
+        item.done = !item.done; 
+      }
+      return item;
+    });
+    this.setState({todos: newTodos});
+  }
+
+  handleRemove = (id) => {
+    const newTodos = [...this.state.todos].filter(item => item.id !== id);
```
isaachinman: Here it's okay to `filter` upon `this.state.todos` directly, as filter doesn't mutate any values, it just returns a new array.

todo-app/src/Components/AddTodoItem.js
```diff
@@ -0,0 +1,99 @@
...
+
+
+   handleAddTodo = () => {
+    const { toBeAddedItem, toBeAddedDeadline } = this.state;
+    this.props.handleAddTodo(toBeAddedItem, toBeAddedDeadline);
+    this.setState({
+        toBeAddedDeadline: '',
+        toBeAddedItem: ''
+    });
+    this.setState({ adding: false })
```
isaachinman: You might want to clear the form here as well? Perhaps it would be worth creating a `clearForm` function to reuse throughout.

todo-app/src/Components/AddTodoItem.js
```diff
@@ -0,0 +1,99 @@
...
+            </div>
+        )
+        }
+    }
+
+    render() { 
+        return (
+            <div>
+                {!this.state.adding && <button className='toggleAddForm' onClick={this.toggleAddFrom} >Add a todo item</button>}
+                {this.showForm()}
```
isaachinman: The setup here is a little confusing. For clarity, I'd suggest moving your condition from inside `showForm` to inline in the JSX to match line 93, like:

    {!this.state.adding && <button className='toggleAddForm' onClick={this.toggleAddFrom} >Add a todo item</button>}
    {this.state.adding && this.showForm()}

This way it's immediately obvious what is going to be rendered in each condition.

todo-app/src/Components/Button.js
```diff
@@ -0,0 +1,35 @@
+//jshint esnext: true
+
+import React, { Component } from 'react'
+
+export default class Button extends Component {
+
+    handleClick = event => {
```
isaachinman: This function isn't used. Did you mean to call `this.props.onClick` inside here?

todo-app/src/Components/todoItem.js
```diff
@@ -0,0 +1,94 @@
...
+    }
+    
+    toggleEditMode = () => {
+        if (!this.props.isEditing) {
+            return (
+                <Button buttonContent='Edit' onClick={this.handleToggleEdit}/>
+            )
+        } else {
+            return (
+                <Fragment>  {/*Remember this is not understood why!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!*/}
```
isaachinman: What do you mean here?

todo-app/src/Components/todoList.js
```diff
@@ -0,0 +1,74 @@
...
+
+    handleChangeDeadline = event => {
+        this.setState({ toBeUpdatedDeadline: event.target.value })        
+    }
+    
+    renderAllTodos = () => {
+        const todoItems = this.props.todos;
+        return todoItems.map(item => {
+            return (
+            <Fragment key={item.id}>
```
isaachinman: Fragments were introduced in React 16 to let you return an array of children. Since you have a single child here, just give the `key` directly to your `TodoItem` component.

todo-app/src/Components/todoItem.js
JiDarwish: Yeah this, since this is being rendered inside a div (in my `render`) Why do I still need to wrap it?? Is it not like only wrap it if it's in render > return???

isaachinman: Ah, I see what you mean. No, the adjacent sibling issue is not specific to the `render` function.

There is nothing special about the `render` function that tells React "this is the place where JSX will be interpreted".

Your `toggleEditMode` function is not returning a string... Your JSX is not "copy/pasted" into place inside your `render` function, it must be interpreted as the function returns.

This question is a little bit outside of the box because returning JSX from class isn't a standard procedure. Of course this is all a matter of opinion, but I would say that if you need a secondary function to return JSX, you should consider creating a new component and put that logic inside the new component's `render` function. To me, the presence of a function like `toggleEditMode` returning JSX indicates that there is probably a component you can abstract out there. Again, it's all opinion.

*The short answer:* the adjacent sibling rule applies wherever JSX is present.

Another option would be to return your children here in an array, but then they need to be comma-separated and have keys:

    toggleEditMode = () => {
        if (!this.props.isEditing) {
            return (
                <Button buttonContent='Edit' onClick={this.handleToggleEdit}/>
            )
        } else {
            return [
              <Button key="toggleEditMode-updateButton" buttonContent='Update' onClick={this.handleUpdate} className='updateButton' />,
              <Button key="toggleEditMode-cancelButton" buttonContent='Cancel' onClick={this.props.handleCancelEdit} className='cancelButton' />
            ]
        }
    }

todo-app/src/App.js
```diff
@@ -1,13 +1,83 @@
...
+      }
+      const oldTodos = this.state.todos || [];
+      
+      this.setState({ todos: [...oldTodos, newItem]})
+    }
+
+  handleToggleDone = (id) => {
+    const newTodos = [...this.state.todos].map(item => {
+      if (item.id === id){
+        item.done = !item.done;
```
joostlubach: This is something from last week, but try to fix it here: in this line, you're *mutating* (changing) an existing todo item. To make this function not mutate, return a new item with the `done` flag inversed. If you're not changing the item (i.e. if `item.id !== id`), you can return the original item.

todo-app/src/App.js
```diff
@@ -1,13 +1,83 @@
...
+        deadline,
+        done: false
+      }
+      const oldTodos = this.state.todos || [];
+      
+      this.setState({ todos: [...oldTodos, newItem]})
+    }
+
+  handleToggleDone = (id) => {
+    const newTodos = [...this.state.todos].map(item => {
```
joostlubach: Same as below: map does not mutate the array, so you don't need to `[...<array>]` it here. Just call `this.state.todos.map(...)`!

todo-app/src/App.js
```diff
@@ -1,13 +1,83 @@
...
+    this.setState({todos: newTodos});
+  }
+
+  handleRemove = (id) => {
+    const newTodos = [...this.state.todos].filter(item => item.id !== id);
+    this.setState({todos: newTodos});
+  }
+
+  handleUpdate = (id, newText, newDeadline) => {
+    const newTodos = [...this.state.todos].map(item => {
```
joostlubach: Same here

JiDarwish:
I've tried to work on all the feedback I got from you and also I fixed the bug about the id's given and added persistency with `localstorage`. I didn't change the names of the props-based functions and the component class functions because some of the functions are being passed down two levels and it's hard to find good names 😄

todo-app/src/App.js
```diff
@@ -1,52 +1,60 @@
 //jshint esnext: true
 
-import React from 'react';
-
-import TodoList from './Components/todoList';
-import AddTodoItem from './Components/AddTodoItem';
-import './App.css';
+import React from "react";
 
+import TodoList from "./Components/TodoList";
```
isaachinman: You've got a compile error here, not sure how it was working on your machine. You're importing from a file that does not exist: `./Components/TodoList`. You just need to capitalise your `todoList.js` file!

todo-app/src/Components/todoList.js
```diff
@@ -1,74 +1,75 @@
-//jshint esnext: true
+import React, { Component } from "react";
 
-import React, { Component, Fragment } from 'react'
-
-import TodoItem from './todoItem';
-import './styles/todoList.css';
+import TodoItem from "./TodoItem";
```
isaachinman: Same here. You just need to capitalise your `todoItem.js` file.

JiDarwish: Hmm that's weird. Cause I got them them both capitalised (also `TodoList.js`) and I pushed them this way, I'm checking `git status` and nothing is different 
See this

todo-app/src/App.js
```diff
@@ -1,13 +1,89 @@
...
+    };
+    const oldTodos = this.state.todos || [];
+
+    this.addNewTodos([...oldTodos, newItem]);
+  };
+
+  handleToggleDone = id => {
+    const newTodos = [...this.state.todos].map(item => {
+      if (item.id === id) {
+        item.done = !item.done;
```
joostlubach: Be sure to create a new `item` here, you're now mutating an existing item, which is not allowed.

todo-app/src/App.js
```diff
@@ -1,13 +1,89 @@
...
+      }
+      return item;
+    });
+    this.addNewTodos(newTodos);
+  };
+
+  handleRemove = id => {
+    const newTodos = [...this.state.todos].filter(item => item.id !== id);
```
joostlubach: Same as above, `this.state.todos.filter(...)` is OK, as `filter` doesn't mutate the array.

todo-app/src/App.js
```diff
@@ -1,13 +1,89 @@
...
+import { loadTodos, setTodos } from "./storage/storage";
+import "./App.css";
+
+export default class App extends React.Component {
+  state = {
+    todos: loadTodos(),
+    isEditing: null
+  };
+
+  addNewTodos = todos => {
```
joostlubach: The name of this method is a bit off, as it doesn't *add* new todos, but rather replaces the list. Perhaps rename to `setTodos` or `updateTodos`?

todo-app/src/Components/AddTodoItem.js
```diff
@@ -0,0 +1,98 @@
...
+        </form>
+      </div>
+    );
+  };
+
+  render() {
+    return (
+      <div>
+        {!this.state.active && (
+          <button className="toggleAddForm" onClick={this.toggleAddForm}>
```
joostlubach: This is more a personal preference, but when you use class names in React, it's a good practice to prefix it with the name of the component, e.g. `"AddTodoItem-toggleAddForm"`

todo-app/src/Components/InputField.js
```diff
@@ -0,0 +1,17 @@
...
+import React, { Component } from "react";
+
+export default class InputField extends Component {
+  render() {
+    return (
+      <input
+        type={this.props.type}
+        placeholder={this.props.placeholder}
+        value={this.props.value}
+        onChange={this.props.handleChange}
```
joostlubach: Small thing to think about: the component `InputField` is there to "wrap" the `<input/>` tag right? That is, if I use `<InputField>` instead of `<input>` I don't have to worry about the DOM.

That's great, but as you're passing `this.props.handleChange` to `onChange`, you expect the code that uses this component to do `event.target.value`, which is DOM-specific.

I would choose to say: `InputField` has a prop `onChange` that takes a callback which will just receive the changed value, and then use it like this: `onChange={event => { this.props.onChange(event.target.value) }}`.

Now if I *use* this component, I can write:

```
<InputField
  value={this.state.myValue}
  onChange={value => this.setState({myValue: value})}
/>
```

See? No more references to DOM-specific stuff!

todo-app/src/Components/todoList.js
```diff
@@ -0,0 +1,75 @@
...
+        <TodoItem
+          key={item.id}
+          data={item}
+          handleToggleDone={this.handleToggleDone}
+          handleRemove={this.props.handleRemove}
+          isEditing={this.props.isEditing === item.id}
+          handleCancelEdit={this.handleCancelEdit}
+          handleToggleEdit={this.handleToggleEdit}
+          handleUpdate={this.handleUpdate}
+          toBeUpdatedItem={this.state.toBeUpdatedItem}
```
joostlubach: You've moved all the state in `TodoList`, where each `TodoItem` has its own little form. I would suggest keeping the form state there, and using a callback only when the editing is complete. That way, you reduce a lot of your code!

todo-app/src/App.js
```diff
@@ -1,13 +1,89 @@
...
+
+  handleToggleEditing = (id, cancel) => {
+    this.setState({ isEditing: cancel ? null : id });
+  };
+
+  handleAddTodo = (text, deadline) => {
+    let newId;
+    if (this.state.todos.length > 1) {
+      let allIds = this.state.todos.map(a => a.id);
+      newId = allIds.reduce((a, b) => Math.max(a, b), 0) + 1; // the max value in the id's + 1
```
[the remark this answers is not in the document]
JiDarwish: Smart! That's super cool

todo-app/src/App.js
JiDarwish: But am I not cloning the array by the `...` spread operator? I mean `newTodos` is supposed to be not the same as my normal `this.state.todos`

isaachinman: @JiDarwish Try to reproduce this as a standalone example. You'll see things aren't quite as you are expecting here.

PR #34, opened 2018-01-10
Eyad-homework3

meyadd:


todo-app/src/App.js
```diff
@@ -1,13 +1,109 @@
...
+    this.setState({tasksArr: updatedtasks});
+	localStorage.tasks = JSON.stringify(updatedtasks);
+  }
+  
+  handleTaskAdd = taskText => {
+	const ids = this.state.tasksArr.map(task => task.id);
+    const newTask =  {
+      id:     ids.length === 0 ? 1 : Math.max(...ids) + 1,
+      description: taskText,
+      deadline:   Moment().format('YYYY-MM-DD'),
```
isaachinman: What happened here? Each task is supposed to have a description and a due date/deadline. Your `deadline` property here is more of a `createdAt` timestamp - there's no way for us to set the deadline to any date other than today.

todo-app/src/App.js
```diff
@@ -1,13 +1,109 @@
...
+      done:   false
+    };
+	const newTaskArr = [newTask, ...this.state.tasksArr];
+	this.setState({tasksArr:newTaskArr});
+    localStorage.tasks = JSON.stringify(newTaskArr);
```
isaachinman: You've got some repetition of `localStorage.tasks` here. Would be nice to see you split this out into a reusable function, perhaps even in a separate file, eg `localStorageUtils.js`.

todo-app/src/App.js
```diff
@@ -1,13 +1,109 @@
...
+import TaskForm from './taskForm';
+import Moment from 'moment';
 
-class App extends Component {
+export default class App extends Component {
+
+  constructor(props) {
+    super(props);
+	let tasks = [];
+	if (localStorage.tasks != null) {
```
isaachinman: Careful here. You should try to use strict equality checks wherever possible. In fact, `localStorage.tasks` will never be `null` unless you set it to `null`, it will be `undefined` by default. The only reason this loose comparison works is because `undefined` and `null` are both falsey.

You can check out the [MDN docs here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined) for some good ideas about how to check against `undefined` in JavaScript.

todo-app/src/App.js
```diff
@@ -1,13 +1,109 @@
...
+		  }
+		  return task;
+		});
+		this.setState({tasksArr: updatedtasks , editingTaskID: null});
+		localStorage.tasks = JSON.stringify(updatedtasks);
+	};
+  
   render() {
+	  
+	if (this.state.tasksArr.length === 0) {
```
isaachinman: Your `if` clause wrapping this JSX is way bigger than it needs to be. In reality you're only switching over your `No items...` vs the `<ul className='task-list'>`. No need to repeat your headers and `TaskForm` component.

I realise you were probably trying to make it clearer exactly what is being returned in what cases, but this would actually get harder to read if the complexity increased any more. Of course this is somewhat a matter of preference, but you can also use the `&&` syntax or a ternary inside a single `return`.

If you don't like that, you can take after @joostlubach's examples with something like:

    return (
      <div>
        <h1>Todo list</h1>
        <h3>Enter description:</h3>
        <TaskForm
          onTaskAdd={this.handleTaskAdd}
        />
        {this.renderTaskList()}
      </div>
    );

Wherein your two view conditions are contained inside the `renderTaskList` function.

todo-app/src/Task.js
```diff
@@ -0,0 +1,96 @@
...
+        this.toggleDone = this.toggleDone.bind(this);
+    }
+
+    toggleDone(event) {
+        const checked = event.target.checked;
+        this.props.onDoneChange(checked, this.props.task.id);
+    }
+	
+	handleDeleteTaskClick = () => {
+		if (window.confirm('Are you sure to delete this task?'))
```
isaachinman: This is a very nice touch!

todo-app/src/App.js
```diff
@@ -1,13 +1,109 @@
 import React, { Component } from 'react';
+import Task from './Task';
+//import tasks from './task.json';
+import TaskForm from './taskForm';
```
isaachinman: Please capitalise file names if they contain a component.

meyadd:
Thanks, I have updated the homework and added the ability to assign a due date and update it

todo-app/src/App.js
```diff
@@ -1,13 +1,95 @@
 import React, { Component } from 'react';
+import Task from './Task';
+import TaskForm from './TaskForm';
```
isaachinman: You have a broken import here. You just need to capitalise your `taskForm.js` file.

todo-app/src/Task.js
```diff
@@ -0,0 +1,118 @@
+import React from 'react';
+import TaskText from './TaskText';
+import TaskDate from './TaskDate';
+import './task.css';
+import LinkButton from './LinkButton';
```
isaachinman: Also a broken impor. You need to capitalise your `linkButton.js` file.

meyadd:
I had changed the file names to capital letter on my pc but it seems the windows system does not distinguish the case of file names and the git push command did not change them on the github server

isaachinman:
Yup, it'd definitely be an OS thing. Good work on the entire project.

PR #35, opened 2018-01-11
Week3

pesh12:
WEEK3
Delete function was already done previously.

But  I have some problem on adding. 
Could you please take a look and help some, 
I feel I almost there but I don't know how to go further

May be my approach completely wrong. haha


todo-app/src/TodoApp.js
```diff
@@ -1,17 +1,64 @@
...
+            className='TextField'
+            value={this.state.commentText}
+            onChange={this.onAddChange}
+            />
+          </div>
+          <div>
+            <button className="add-button" onClick={this.onAddClick}> Add</button>
+          </div>  
+        </div>
         <TodoList/>
```
isaachinman: Hint...

todo-app/src/TodoApp.js
```diff
@@ -0,0 +1,64 @@
+import React, { Component } from 'react';
+import TodoList from './TodoList';
+import todoItems from './todo.json';
+
+export default class TodoApp extends Component {
+
+  state = {
+    commentText: '',
```
joostlubach: be careful to name your variables properly! This is a todo app, not the blog app.

todo-app/src/TodoApp.js
pesh12: Unfortunately, I couldn't get it. @isaachinman

pesh12: I could make it, if I could set the state in TodoList. 
I could't figure out how to carry the state down to the tree.

todo-app/src/TodoApp.js
pesh12: Changed 👍  @joostlubach 
maybe it was because  watching the video a lot :))

todo-app/src/TodoApp.js
isaachinman: @pesh12 Yes, you have the right idea.

The problem is that in both `TodoApp.js` and `TodoList.js` you are loading data from the json file into state. This means that even when `TodoApp` changes its todos, `TodoList` doesn't care as it has its own version of the list in internal state.

You need to pass `TodoApp`'s state into `TodoList` like this:

    <TodoList todos={this.state.todos} />

And then wherever you are referencing `this.state.todos` inside `TodoList`, simply change to `this.props.todos`.

Let me know how it goes...

pesh12: I made a function in TodoList, to handle the "statechange"

It did not change the state, but changed the props (adds the value in TodoList). 
But I put a console.log in the function, it never works.
I think I am really confused :))
We may sit on Sunday and see the problem together.

pesh12:
now the program adds. Now I need to focus on "this week's tasks". 
Thanks for helping

PR #36, opened 2018-01-12
Homework week3

Elias011:


todo-app/src/App.js
```diff
@@ -1,13 +1,98 @@
...
+            return {...todoItem, description};
+        }else {
+            return todoItem;
+        }
+    }))
+};
+toggleStatus = (todoID) =>{
+    this.setTodos(this.state.todos.map(todo =>{
+                if (todo.id === todoID){
+                    todo.done = !todo.done
```
isaachinman: You are mutating your todo object directly here. Instead, you need to create a new object, like you do inside your `handleSaveEdit` function just above.

todo-app/src/App.js
```diff
@@ -1,13 +1,98 @@
...
+     };
+ }
+ handleTaskAdd = newTodo => {
+     const todo = this.newTaskWithText(newTodo);
+     this.setTodos([todo, ...this.state.todos]);
+ };
+
+handleRemoveTodo = (todoItem) => {
+     const updateTodos = this.state.todos;
+    updateTodos.splice(updateTodos.indexOf(todoItem), 1)
```
isaachinman: It would be much more robust to have a `handleRemoveTodo` function that simply took a todo's ID and removed it via `filter`. What you're doing here is actually forcefully mutating your state before even calling `setState`, a React no-no.

Even though you are doing `const updateTodos = this.state.todos`, you're not actually creating a new array anywhere, and `splice` modifies an array in place instead of returning a new array.

todo-app/src/App.js
```diff
@@ -1,13 +1,98 @@
...
       <div>
-        Hello World!
+        <h1>Todo App</h1>
+         <TaskForm 
+           ontodoAdd = {this.handleTaskAdd}
+          />
+		<ul className='tasks'>
+         {todos.map((todoItem, i) =>(
+        <Task 
+        key = {todoItem.id}
```
isaachinman: Just a head's up, since your todo IDs are simple numbers, you might want to make your keys here something like `task-${todoItem.id}`, etc. Keys need to be unique across an entire app (or at least the current view), so if you had other items with numerical IDs, you might see a clash.

todo-app/src/Task.js
```diff
@@ -0,0 +1,46 @@
...
+import './index.css';
+
+export default class Task extends React.Component{
+
+  render() {
+  const{todoItem, toggleStatus, isEditing, onRemove, onEdit, onCancleEdit, onSaveEdit} = this.props;
+       return (
+            <li className="task-list">
+             <span> <input type = 'checkbox' checked = {todoItem.done} onChange = {() => toggleStatus(todoItem.id)}/></span>
+            <span> {todoItem.done? <span className = 'task-not-done'>
```
isaachinman: You have something strange going on here. Why are you giving a class of `task-not-done` for tasks that _are_ done? I see this CSS declaration:

    .task-not-done {
      text-decoration:line-through;
    }

Seems you might have just named it incorrectly?

PR #37, opened 2018-01-17
Eyad-homework4

meyadd:
**todo list** app by MobX stores (without using provider and inject)


isaachinman:
Hi @meyadd, that was fast! There's a problem before we can get started though:

You did not commit the `scripts` directory which `create-react-app` outputs on `eject`. We need this directory to be able to run your app.

meyadd:
hi @isaachinman , 
I have committed the  `scripts `directory

isaachinman:
Hi @meyadd, we're still missing some key files. Now I get the error:

> Error: Cannot find module '../config/env'

In a more "formal" or "production" environment, your app would be building inside a container, so these sorts of things would not pass CI.

In this case, you can try cloning your repo into a new directory on your machine, do a `yarn install`, and then try `yarn start`.

meyadd:
sorry @isaachinman,
I have committed the missed files and then I have tested the project as you said, everything work fine now

todo-app/src/TasksStore.js
```diff
@@ -0,0 +1,60 @@
...
+
+    @computed
+    get doneCount() {
+            return this.tasks.filter(task => task.done).length;
+    }
+}
+
+const tasksStore = new TasksStore();
+
+export default tasksStore;
```
isaachinman: Just so you know, you can do this in one line:

    export default new TasksStore();

todo-app/src/App.js
```diff
@@ -1,13 +1,56 @@
...
+	handleTaskStartEdit = taskId => {
+		this.setState({editingTaskID: taskId});
+	};
+	
+	handleTaskStopEdit = () => {
+		this.setState({editingTaskID: null});
+	};
+	
+	
+
```
isaachinman: No more than two consecutive empty lines, please.

todo-app/src/TasksStore.js
```diff
@@ -0,0 +1,60 @@
...
+    @action
+    changeDone(doneValue, taskId) {
+        const task = this.findTask(taskId);
+        if (task == null) { return }
+        task.done = doneValue;
+    }
+
+    @computed
+    get doneCount() {
+            return this.tasks.filter(task => task.done).length;
```
isaachinman: Please mind your indentation.

todo-app/src/App.js
```diff
@@ -1,13 +1,56 @@
...
+	  	{tasksStore.tasks.length === 0 ? 
+	  		<div>No items...</div> 
+	  	: 
+				<ul className='task-list'>
+					{
+						tasksStore.tasks.map(item => 
+						<Task 
+							key={item.id} 
+							id={item.id} 
+							tasksStore = {tasksStore}
```
isaachinman: Because you are only using actions from the store inside your `Task` component, it might be nice to explicitly pass just those actions to `Task` so that it is clear what its requirements are.

MobX does do intelligent observation based on store values, but regardless you are still injecting a potentially large amount of useless data into each `Task` element on the page - imagine if we had a few hundred or thousand.

todo-app/src/TaskForm.js
```diff
@@ -0,0 +1,61 @@
...
+export default class TaskForm extends React.Component {
+
+  state = {
+    taskText: '',
+    taskDate: Moment()
+  };
+
+  handleAddTaskClick = () => {
+    this.props.tasksStore.addTask(this.state.taskText , this.state.taskDate.format('YYYY-MM-DD'));
+    this.setState({taskText: '' , taskDate: Moment()});
```
isaachinman: You do have duplication of state declaration here, although it's only twice. If you had a third or fourth instance, you could start to think about doing something like this:

    const initialState = {
      taskText: '',
      taskDate: Moment()
    };

    @observer
    export default class TaskForm extends React.Component {

      state = { ...initialState }

      handleAddTaskClick = () => {
        this.props.tasksStore.addTask(this.state.taskText , this.state.taskDate.format('YYYY-MM-DD'));
        this.setState({ ...initialState });
      };

PR #40, opened 2018-01-20
Week4 

pesh12:
Sorry for late. I had very busy week.


todo-app/src/TodoItem.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { observer } from "mobx-react";
+import Description from './Description';
+import Deadline from './Deadline';
+import './index.css';
+
+
+@observer
```
joostlubach: Is it necessary to make the `TodoItem` component a MobX-observer? `TodoItem` is not getting its data directly from the store, only from its props, so you don't need to do this.

todo-app/src/TodoList.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import TodoItem from './TodoItem';
+import {observer} from 'mobx-react';
+import {todoStore} from './stores';
+
+@observer
```
joostlubach: Don't forget: in the recent version of `mobx-react` you need to also add `@inject`.

todo-app/src/TodoList.js
```diff
@@ -0,0 +1,67 @@
...
+
+  onAddChange = event => {
+    var text = event.target.value;
+    this.setState({
+      text: text
+    });
+  };
+
+  onDelete = (todoID) => {
+    todoStore.onDelete(todoID);
```
joostlubach: Small thing: `todoStore.onDelete` is not an event handler, but a method on `todoStore`. Best to just call it `delete` or `deleteTodo`.

todo-app/src/TodoList.js
```diff
@@ -0,0 +1,67 @@
...
+      text: text
+    });
+  };
+
+  onDelete = (todoID) => {
+    todoStore.onDelete(todoID); 
+  } 
+  
+  onDone = (todoID) => {
+   todoStore.onDone(todoID);
```
joostlubach: Same here

todo-app/src/TodoList.js
```diff
@@ -0,0 +1,67 @@
...
+            onChange={this.onAddChange}
+            />
+          </div>
+          <div>
+            <button className="add-button" onClick={this.onAddClick}> Add</button>
+          </div>  
+        </div>
+          <ul>
+            {todoStore.todos.map((todoItem, i) => (
+              <React.Fragment key={todoItem.id}>
```
joostlubach: No need to use a `Fragment` here, as there's only one child. You can just render the `<TodoItem>` directly.

todo-app/src/stores/TodoStore.js
```diff
@@ -0,0 +1,61 @@
...
+    onDone(todoID){
+
+      const todo = this.todos.find(todo => todo.id === todoID)
+      if(todo == null) {return}
+
+      todo.done = !todo.done
+    }
+
+    @action
+    onDelete(todoID){
```
joostlubach: See comment on naming above.

todo-app/src/stores/TodoStore.js
```diff
@@ -0,0 +1,61 @@
...
+        "description": text,
+        "deadline": "2018-01-11",
+        "done": false
+      }
+
+      this.todos.push(todo);
+    }
+
+    @action
+    onDone(todoID){
```
joostlubach: See comment on naming above.

todo-app/src/stores/TodoStore.js
```diff
@@ -0,0 +1,61 @@
...
+        "id": ids.length === 0 ? 1 : Math.max(...ids) + 1,
+        "description": text,
+        "deadline": "2018-01-11",
+        "done": false
+      }
+
+      this.todos.push(todo);
+    }
+
+    @action
```
joostlubach: Great! This is an action method, and you're correct to mark it as such.

todo-app/src/TodoItem.js
pesh12: I think I need that, 'deleteTodo' and 'changeStatus' do not functioning otherwise.

todo-app/src/TodoList.js
pesh12: Added :)

todo-app/src/TodoItem.js
isaachinman: See #38 and Ji's [detailed explanation here](https://hackyourfuture.slack.com/archives/C6CK8SF7S/p1516371508000559).

PR #41, opened 2018-01-21
homework week 5

JiDarwish:
Hey, here's the homework for this week. One small issue, I downloaded an image and it wouldn't display it! No idea why TodoList.js line 29 there is a big comment about it.
Thanks!


todo-app/src/Components/todoList.js
```diff
@@ -0,0 +1,41 @@
...
+          data={item}
+          handleToggleDone={this.handleToggleDone}
+        />
+      );
+    });
+  };
+
+  render() {
+    if (this.props.todoStore.loading) {
+      // return <img src="./loading.png" alt="loading" className="loadingImg" />; ////////////////////this is Not working WHY?
```
isaachinman: Okay, fair question. The basic reason this doesn't work is that your `loading.png` image is inside your `src` dir, which is never exposed publicly. We aren't working with an old-school `index.html` with some css and images next to it in an exposed directory.

For CRA, and all modern projects, JavaScript, CSS, images, and pretty much any other asset you can think of are going to be handled, preprocessed, or transpiled via a build tool (typically Webpack at this moment in time).

That being said, your `loading.png` image _definitely_ should be inside your `src` dir as it's part of your "source code", despite being an image. Here's how:

    import loaderImage from './loading.png'
    loading

Now, do this and then inspect the image on the browser page. Look at the `src` attribute. You'll see that there's a lot going on behind the scenes here.

todo-app/src/Components/AddTodoItem.js
```diff
@@ -0,0 +1,101 @@
...
+        <form onSubmit={this.handleSubmit}>
+          <div className="formInputWrapper">
+            <InputField
+              type="text"
+              placeholder="Todo item"
+              className="todoInput"
+              value={this.state.toBeAddedItem}
+              handleChange={this.handleChangeTodoItem}
+            />
+            <InputField
```
isaachinman: It would be nice to have a datepicker here instead of a text input. At the moment it's very easy to type an invalid date in here and break the app (no error feedback, by the way). There is a [native HTML5 date input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date) (although it's famous for being terrible). There are also many, many third-party datepicker components for React.

todo-app/src/Components/TodoItem-newTodoFormToggle.js
```diff
@@ -0,0 +1,62 @@
...
+        </Fragment>
+      );
+    }
+    const { description, deadline, _id } = this.props.data;
+    const done = this.props.done;
+    const inputId = `item${_id}`;
+    return (
+      <Fragment>
+        <div className="todoDate">
+          <span>{moment(deadline).format("YYYY-mm-DD hh:mm")}</span>
```
isaachinman: You have a bug here in your formatting. Values are case sensitive.

todo-app/src/store/TodoStore.js
```diff
@@ -0,0 +1,126 @@
...
+
+useStrict(true);
+// will throw an error if you didn't use the @action or didn't use any decorator
+
+export default class TodoStore {
+  @observable todos = [];
+
+  @observable loading = false;
+
+  @observable isEditing = null;
```
isaachinman: This is now confusing. You have `loading = false`, and then immediately below you have `isEditing = null`. To be more accurate, and to make sense to future devs reading the code, you should probably change this to something like `isEditingID`.

PR #42, opened 2018-01-23
homework week4

Elias011:
Sorry for late


isaachinman:
You project is broken. Try cloning your PR from scratch, doing `yarn` and then `yarn start`. You need to commit certain files for this to work.

Elias011:
Sorry for the mistake

todo-app/src/stores/TaskStore.js
```diff
@@ -0,0 +1,52 @@
...
+            done: false
+        }
+       this.todos.push(todo);  
+    }
+    @action
+      removeTask(id) {
+      this.todos = this.todos.filter(task => task.id !== id)
+  }
+     @action
+     toggleStatus(todoID) {
```
isaachinman: What is going on in this function? You are mapping through `this.todos` on your store and assigning that new array to a value of `todo`, and then returning `todo` _inside_ your map block, which is undefined.

So your `const todo` ends up equaling `[undefined, undefined, undefined]`.

It appears it's completely an accident that this function works. You are mutating the specific value directly inside your `map` function.

You have two options:

1. Mutate the specific todo directly without doing this unnecessary `map` and assigning an array of `undefined`s to a `const`. This would look like:

>const todoToToggle = this.todos.filter(todo => todo.id === todoID)[0]
> // Now reverse the value

2. Return a new array properly, without mutating your old data, and set `this.todos` to it.

I _highly_ recommend attempting the second option. It seems like you already had this in mind, you just lost track halfway through.

The first clue is that `todo` is of course undefined inside your `map` block.

todo-app/src/stores/TaskStore.js
```diff
@@ -0,0 +1,52 @@
...
+                if (todoItem.id === todoID){
+                    todoItem.done = !todoItem.done
+                }
+                return todo
+                
+            })
+    }
+
+     @action
+     handleSaveEdit(id , description){
```
isaachinman: You have the same problem here as with your `toggleStatus` action.

todo-app/src/stores/TaskStore.js
```diff
@@ -0,0 +1,52 @@
...
+      saveTodoItems(this.todos);
+    });
+  }
+    
+    @observable
+    todos = loadTodoItems();
+    
+
+    @action
+    addTask(text){
```
isaachinman: You refer to your "things" as both "todos" and "tasks". It would be less confusing if you chose one term and stuck with it.

todo-app/src/storage.js
```diff
@@ -0,0 +1,8 @@
+export function loadTodoItems() {
+   if(localStorage.todos == null){ return [] }
```
isaachinman: Careful here. You should try to use strict equality checks wherever possible. In fact, `localStorage.tasks` will never be `null` unless you set it to `null`, it will be `undefined` by default. The only reason this loose comparison works is because `undefined` and `null` are both falsey.

You can check out the [MDN docs here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined) for some good ideas about how to check against `undefined` in JavaScript.

todo-app/src/Task.js
```diff
@@ -0,0 +1,47 @@
...
+@inject('taskStore') 
+@observer
+export default class Task extends React.Component {
+
+
+  render() {
+      const{todoItem, toggleStatus, isEditing, onRemove, onEdit, onCancleEdit, onSaveEdit} = this.props;
+        return (
+            <li className="task-list">
+             <span> <input type = 'checkbox' checked = {todoItem.done} onChange = {() => toggleStatus(todoItem.id)}/></span>
```
isaachinman: I cannot even read this file due to formatting. Please reformat with proper HTML indentation.

PR #43, opened 2018-01-24
Eyad-homework5

meyadd:


todo-app/src/App.js
```diff
@@ -1,13 +1,63 @@
...
+import tasksStore from './TasksStore';
+import {observer} from 'mobx-react';
+
+@observer
+export default class App extends Component {
+
+  constructor(props) {
+		super(props);
+	
+    this.state = {
```
isaachinman: Sorry, I guess we missed this last week... did you have a reason for leaving this state inside a React component instead of putting it inside your MobX store? I think it would be better to move it inside the store.

todo-app/src/App.js
```diff
@@ -33,10 +36,13 @@ export default class App extends Component {
 			<h4 className='task-done'>✓ = {tasksStore.doneCount}</h4>
 			<h3>Enter description:</h3>
 			<TaskForm tasksStore = {tasksStore} />
-	  	{tasksStore.tasks.length === 0 ? 
-	  		<div>No items...</div> 
-	  	: 
-				<ul className='task-list'>
+	  	{tasksStore.loadingTasks ? (
```
isaachinman: When you have nested ternaries like this it starts to get pretty hard to read. Can you refactor this into multiple returns with `if` statements so that it's more clear?

todo-app/src/TasksStore.js
```diff
@@ -1,52 +1,136 @@
...
         return this.tasks.find(task => task.id === taskId)
     }
 
+    @action
+    loadTasks(showLoader = true){
+        if (showLoader) {
+            this.loadingTasks = true
+        }
+        fetch(`${api_url}/todos`)
+        .then(res =>{return res.json()})
```
isaachinman: With arrow functions, you can actually skip the block declaration and `return` keyword if you're doing everything in one line:

    .then(res => res.json())

From the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions):

> Arrow functions can have either a "concise body" or the usual "block body".
    In a concise body, only an expression is specified, which becomes the explicit return value. In a block body, you must use an explicit return statement.

todo-app/src/TasksStore.js
```diff
@@ -1,52 +1,136 @@
...
+        .then(tasks_arr => {
+            runInAction(()=>{
+                this.tasks = tasks_arr.map(comment => {
+                    return {
+                        id:comment._id ,
```
isaachinman: So, the only property you are actually changing is `id` (from `_id`).

If the object was very large, or if you did not care about retaining the original `_id` property, you could also do:

    return {
      id: comment._id,
      ...comment,
    }

meyadd:
hi @isaachinman 
thanks for your feedback, I have updated my code

PR #44, opened 2018-01-24
Amin week3

aminmahrami:


todo-app/src/Time.js
```diff
@@ -0,0 +1,16 @@
...
+
+
+export default class Time extends React.Component{
+    
+    render(){
+        return (
+
+            <div>
+                {this.props.title}
+                {this.props.time}
```
joostlubach: It would be nice if the `Time` component could format the time properly. Currently, the date is shown as it was entered, but perhaps you can use `momentjs` to show it nicer?

todo-app/src/Todo.js
```diff
@@ -0,0 +1,129 @@
...
+          <div>
+            <div>
+              {/* this is for task */}
+              <TextField
+                multiline
+                value={this.state.TodoText}      
+                onChange={this.handleTextChange} 
+              />
+              {/* this is for deadline date */}
+              <TextField value={this.state.TodoText}
```
joostlubach: If this is for the deadline, why is its value set to `{this.state.TodoText}`?

todo-app/src/TextField.js
```diff
@@ -0,0 +1,51 @@
...
+
+export default class TextField extends React.Component {
+
+    render() {
+        if (this.props.multiline) {
+            return this.renderMultiLine();
+        } else if (this.props.singleline) { 
+            return this.renderSingleLine();
+        } else {
+            return this.renderDateForm();
```
joostlubach: Your component is called `TextField`, but you've made it so that it also allows entering dates. It makes more sense to create a new component called `DateField` and use that instead. Now it's sort of confusing that if I use `<TextField/>` it's actually for a date, and if I use `<TextField singleline/>` it's a text field.

todo-app/src/TextField.js
```diff
@@ -0,0 +1,51 @@
...
+        );
+    };
+
+    renderDateForm() {
+        return (
+            <div>
+                <input type="date" 
+                className="date"
+                value={this.props.value}
+                onChange={this.props.onChange}
```
joostlubach: If I enter '2018-01-27' in the field, is `onChange` called with a `Date` or with a `string`? And what would you want to get out of the field, a `Date` or a `string`?

todo-app/src/Todo.js
```diff
@@ -0,0 +1,129 @@
...
+
+    handleDeadlineChange = event => {
+        this.setState({deadline: event.target.value});
+    };
+
+    handleSaveClick = () => {
+        const {todoText, deadline } = this.state;
+        this.props.onSaveClick(todoText);
+        this.setState({todoText: ''});
+        this.props.onSaveDateClick(deadline);
```
joostlubach: Why do you have two `onSave...` methods? If you always call them together, why not have one single `onSave` method which passes an entire `todo` object?

todo-app/src/Todo.js
```diff
@@ -0,0 +1,129 @@
...
+                onChange={this.handleDeadlineChange} 
+            
+              />
+            </div>
+            <div className="edit-mode-buttons">
+              <Button
+                label='Save'
+                onClick={this.handleSaveClick}
+                //disabled is not working at this momment
+                //disabled={this.state.TodoText.trim() === ''}
```
joostlubach: Do you know why it's not working?

todo-app/src/TodoApp.js
```diff
@@ -0,0 +1,112 @@
...
+              return {...todo, task};
+            } else {
+              return todo
+            }
+      }));
+    };
+
+    //saves deadline after editing
+    //both handleSave and handleSaveDeadline are not working at  
+    //sametime thats why i've used setTimeOut until updating to mobx
```
joostlubach: Typically if you have to resort to such measures, you have some other problem in your code. In this case, why not have one single `handleSave` method, which can update multiple properties at once?

joostlubach: The reason why it doesn't work in two goes, is because `setState` updates the state in the next cycle (that is, until JavaScript has finished executing all code). You should not call `setState` and then later use `this.state....` because you will still use the old version. 

But, all of that is not a problem if you would use one function to update multiple properties.

PR #45, opened 2018-01-25
Amin week4

aminmahrami:


todo-app/src/Title.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+export default class Title extends React.Component {
+    
+    render(){
+        return(
+            <span 
+            disabled={this.props.disabled}
```
joostlubach: What would you expect to happen if you disable a `<span>` element?

todo-app/src/TodoList.js
```diff
@@ -0,0 +1,66 @@
...
+  handleEditTodo = (id) => {
+    this.setState({editingTodoID: id});
+  };
+  
+  handleTodoCancelEdit = () => {
+    this.setState({editingTodoID: null})
+  };
+
+  handleTodoSave = (id, text) => {
+    this.props.todos.onSaveTodo(id, text);
```
joostlubach: Small thing: I would name it `saveTodo`, as it's not an event handler, but simply a method on the store.

todo-app/src/stores/TodosStore.js
```diff
@@ -0,0 +1,67 @@
...
+                task: text,
+                addedTime: Moment().format('YY-MM-DD'),
+                deadlineTime: deadline,
+                done: false,
+        };
+        this.todos.push(todo);
+    };
+
+    @action
+    handleTodoRemove = (id) => {
```
joostlubach: Also here, it's not an event handler, but just a method.

todo-app/src/stores/TodosStore.js
```diff
@@ -0,0 +1,67 @@
...
+              return {...todo, task};
+            } else {
+              return todo
+            }
+      });
+    };
+
+    @action
+    onSaveDeadline = (id, deadlineTime) => {
+        this.todos = this.todos.map(todo => {
```
joostlubach: MobX doesn't have the same requirements as React about mutability. You can just find the todo and do `todo.deadline = deadlineTime` here!

PR #47, opened 2018-01-26
Week5

pesh12:
Dear Joost and  Isaac. 
I really don't know how to thank to you.
What you are doing is very precious.
We couldn't make it without HYF and people like you.

Thanks


isaachinman:
I get this error:

screen shot 2018-01-26 at 16 25 25

You need to install `babel-plugin-transform-decorators-legacy` as a dependency.

isaachinman:
And no need to thank us, it is our pleasure!

pesh12:
I have also forked my work to another directory to see the mistake. As you mentioned, I have installed babel-plugin as dependency, now it works. 
And in case;  I have installed "moment" as dependency as well.

todo-app/src/todo.json
```diff
@@ -0,0 +1,20 @@
+[
```
isaachinman: This file can be removed now.

todo-app/src/TodoList.js
```diff
@@ -0,0 +1,68 @@
...
+    todoStore.onDelete(todoID); 
+  } 
+  
+  changeStatus = (todoID) => {
+   todoStore.onDone(todoID);
+  };
+
+
+  render() {
+    if(todoStore.todos.length === 0){
```
isaachinman: When I load the page, it says "No items..." while the todos are loaded from the API. This is not a great user experience.

todo-app/src/stores/TodoStore.js
```diff
@@ -0,0 +1,108 @@
...
+    // setTodos(todos) {
+    //   this.todos = todos
+    // }
+
+    @action  // add a todo item
+    addTodo(text){
+
+      const todo = {
+        "description": text,
+        "deadline": Moment().format('YYYY-MM-DD'),
```
isaachinman: First of all, these keys do not need quotes. This is JavaScript, not JSON.

Second, and more importantly, your application has _no way_ to actually set the deadline. This means your application can only create todos that are due today! Feel free to ask some of the other guys for examples. You will need to build this functionality in.

todo-app/src/stores/TodoStore.js
```diff
@@ -0,0 +1,108 @@
+import {observable, action, useStrict, runInAction} from 'mobx'
+import Moment from 'moment'
+
+
+useStrict(true)
+
+export default class TodoStore{
+
+    constructor(){
+      this.loadTodos();   // without constructor it won't work, you need to use componentDidMount
```
isaachinman: What does this comment mean? Why are you not calling `loadTodos` in `componentDidMount`?

todo-app/src/stores/TodoStore.js
```diff
@@ -0,0 +1,108 @@
...
+      })
+        .then(response => response.json())
+        .then(response => {
+          console.log("POST RESPONSE   :", response)
+          this.loadTodos();
+        })  
+    }
+
+    @action // change status of a todo
+    onDone(todoID){
```
isaachinman: I know that you simply copy/pasted these functions from your original React component, but you will need to be more careful. First, this is no an "on" handler. It's not a React event handler, it's a MobX action within a store. Second, it is not only associated with being "done", it toggles the completion status back and forth.

todo-app/src/stores/TodoStore.js
```diff
@@ -0,0 +1,108 @@
...
+
+      const todo = this.todos.find(todo => todo._id === todoID)
+      if(todo == null) {return}
```
isaachinman: Don't write code like this, it's not helpful. First of all, use strict equality checks. Second, if something goes wrong, you should throw an error instead of silently returning the function. The whole point of creating safeguards like this is to create maintainable and debuggable code.

todo-app/src/stores/TodoStore.js
```diff
@@ -0,0 +1,108 @@
...
+          console.log("PATCH RESPONSE   :", response)
+          this.loadTodos();
+        })  
+    }
+
+    @action  // delete a todo item
+    onDelete(todoID){
+
+      var updatedTodos = this.todos.filter((val,index) =>{
+      return val._id !== todoID;
```
isaachinman: Watch your indentation, please.

todo-app/src/stores/TodoStore.js
```diff
@@ -0,0 +1,108 @@
...
+        .then(response => {
+          console.log("PATCH RESPONSE   :", response)
+          this.loadTodos();
+        })  
+    }
+
+    @action  // delete a todo item
+    onDelete(todoID){
+
+      var updatedTodos = this.todos.filter((val,index) =>{
```
isaachinman: Where did this `var` come from? Please use `const`.

Also, your `index` is unused here and should be removed.

todo-app/src/stores/TodoStore.js
pesh12: Actually it was for myself to future check, 
I meant that if you won't use constructor you have to use "ComponentDidMount" utility.
In class you did it in both ways, and I didn't want to miss any of them :)
I know I will need to remember all the stuff

todo-app/src/stores/TodoStore.js
pesh12: Actually, I haven't used the Moment before.
I just wanted to practice/ apply.
Now, I have changed the deadline as always 4 days later.
Moment is great!!

todo-app/src/stores/TodoStore.js
pesh12: I have changed the names of the functions
onDone    to   todoStatus
onDelete  to   deleteTodo

pesh12:
Dear Isaac @isaachinman , 
I have changed the code according to your suggestions.
Thanks for your time.

todo-app/src/stores/TodoStore.js
isaachinman: Ah okay, got it. As Joost pointed out in class, it's probably better to use `componentDidMount`, so that the lifecycle of the data is directly related to the lifecycle of the UI.

isaachinman:
Hi @pesh12, and thanks!

This isn't your best work, though... You simply changed the deadlines to all be four days in the future. That's also not a very useful app, one that only allows todos to be due four days from now. It's no better than an app which only allows todos to be due today.

Also, for the loading situation, you simply changed it to:

    if(todoStore.todos.length === 0){
      return(
        <div className="TodoList">Loading...</div>
      )
    }

Again, this is not any better than what you previously had. Sure, now it seems OK when we are actually loading todos. But what if there are zero todos?

PR #48, opened 2018-01-26
Amin week5

aminmahrami:


todo-app/src/Todo.js
```diff
@@ -0,0 +1,135 @@
...
+
+
+    renderEditing() {
+        return (
+          <div>
+            <div>
+              {/* this is for description */}
+              <TextField
+                multiline
+                value={this.props.description}
```
isaachinman: I guess you did not test your app very thoroughly.

The todo descriptions and dates are not editable whatsoever. Your problem lies here, you are updating state with `this.handleTextChange`, but passing `this.props.description` to the field as a value.

todo-app/src/App.js
```diff
@@ -1,13 +1,19 @@
...
-class App extends Component {
-  render() {
-    return (
-      <div>
-        Hello World!
-      </div>
-    );
-  }
-}
+window.stores = stores;
```
isaachinman: This does not belong in "production" code. If you use something like this for testing, make sure to remove it before committing. Also, the MobX devtools might suit you better. That being said I think you just copy/pasted it from Joost's example.

todo-app/src/TextField.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+
+
+export default class TextField extends React.Component {
```
isaachinman: This needs to be two components, not one. It's fine to render either a textarea or a single-line input from a component with configurable props, but rendering a date input from a component called `TextField` does not make any sense.

todo-app/src/Todo.js
```diff
@@ -0,0 +1,135 @@
...
+                <div className="Todo-Description">
+                    <Description 
+                    description={todo.description}
+                    />
+                </div> 
+                
+                <div className="Add-Time">
+                    <div>
+                        <div className="Add-Deadline-Title">
+                           <Title disabled={this}
```
isaachinman: What is this component? What is its purpose? Does this really need to be a component? What is this "disabled" prop doing?

todo-app/src/Title.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+export default class Title extends React.Component {
+    
+    render(){
+        return(
+            <span 
+            disabled={this.props.disabled}
```
isaachinman: Why did you put `disabled` on a `span`?

todo-app/src/Todo.js
```diff
@@ -0,0 +1,135 @@
+import React from 'react';
+import {inject, observer} from 'mobx-react';
+import Description from './Description';
+import Time from './Time';
+import Title from './Title';
+import Button from './button';
+import TextField from './TextField';
+
+@inject('todos')@observer
```
isaachinman: It's conventional to have line breaks between decorators.

todo-app/src/Todo.js
```diff
@@ -0,0 +1,135 @@
...
+                
+                <div className="Add-Time">
+                    <div>
+                        <div className="Add-Deadline-Title">
+                           <Title disabled={this}
+                            titleText={"Deadline is "}
+                            />
+                        </div>
+                        <div className="Add-Deadline-Time">
+                            <Time
```
isaachinman: Again, what is the purpose of this component? It just renders text into a div...

todo-app/src/TodoList.js
```diff
@@ -0,0 +1,73 @@
+import React from 'react';
+import {inject, observer} from 'mobx-react';
+import Todo from './Todo';
+
+@inject('todos')@observer
+export default class TodoList extends React.Component {
+
+  state = {
+    editingTodoID: null,
```
isaachinman: I think it would be sensible to move this `editingTodoID` into your MobX store. You currently have state in at least four places:

1. MobX store
2. Here in `TodoList`
3. In `TodoForm` for creating new todos
4. In `Todo` for modifying existing todos

Let's see if you can lift some of this state out of React components and into your MobX store.

todo-app/src/data/todo-data.json
```diff
@@ -0,0 +1,18 @@
+[{
```
isaachinman: This should be removed now.

todo-app/src/storage.js
```diff
@@ -0,0 +1,8 @@
+export function loadTodos() {
```
isaachinman: This should also be removed - you are persisting data via the API, not localStorage.

todo-app/src/stores/TodosStore.js
```diff
@@ -0,0 +1,95 @@
...
+
+export default class TododStore {
+
+    @observable
+    todos = [];
+  
+    loadTodos(){
+    fetch("https://hyf-react-api.herokuapp.com/todos")
+        .then(response => response.json())
+        .then(response => {
```
isaachinman: What happens in the case of an error? You have no `catch` here.

todo-app/src/stores/TodosStore.js
```diff
@@ -0,0 +1,95 @@
...
+            headers: {
+            'Content-Type': 'application/json',
+            }, body: JSON.stringify({description : text}),
+        })
+        .then(response => response.json())
+        .then(response => this.loadTodos())
+    };
+
+    @action
+    onSaveDeadline = (id, deadlineTime) => {
```
isaachinman: Why do you have two PATCH functions, one for `description` and one for `deadline`? Can you figure out a way to instead have a single `updateTodo` function?

todo-app/src/stores/TodosStore.js
```diff
@@ -0,0 +1,95 @@
...
+            headers: {
+            'Content-Type': 'application/json',
+            }, body: JSON.stringify({deadline : deadlineTime}),
+        })
+        .then(response => response.json())
+        .then(response => this.loadTodos())
+    };
+
+    @action
+    handleDoneToggle = (id) => {
```
isaachinman: There is no need for any `handle` naming here. These are no longer React event-handlers, they are MobX actions.

PR #46, opened 2018-01-26
Week 5 including week 3 and 4

itemah26:

PR #49, opened 2018-01-27
homework wee5 

Elias011:
I have fixed my the issues of homework week 4
I also tried to make it less confusing and I hope that the formatting is now  better than before


isaachinman:
Now seeing this error:

screen shot 2018-01-28 at 20 52 35

Same issue as in #46.

Elias011:
Sorry Isaac!! I forgot to --save 'Moment' in my package.json

todo-app/src/stores/TodoStore.js
```diff
@@ -0,0 +1,110 @@
...
+      .then(res => res.json())
+      .then(res => {
+        this.loadTodoItems(false)
+      })
+
+  }
+
+  @action
+
+  handleSaveEdit(id, description) {
```
isaachinman: First, the decorator should not have a line break after it.

Second, you have like 4-5 declarations of a `handleSaveEdit` function in this codebase, which is confusing.

Third, this is a MobX store, not a React component. You should not use the `handle` naming as this is not an event hander at all, it is a MobX action.

todo-app/src/TodoList.js
```diff
@@ -0,0 +1,74 @@
...
+import TodoForm from './TodoForm';
+import loadinImage from './loadingImage.gif';
+
+@inject('todoStore')
+@observer
+
+export default class TodoList extends React.Component {
+
+  state = {
+    editingTodoID: null
```
isaachinman: Why is this state not inside your MobX store? Please move it inside your MobX store unless you can think of a good reason why it should not be there.

todo-app/src/TodoList.js
```diff
@@ -0,0 +1,74 @@
...
+        <img src={loadinImage} alt='loading' />
+      )
+    }
+    return (
+      <div>
+        <TodoForm
+          ontodoAdd={this.handleTodoAdd}
+        />
+        <ul className='tasks'>
+          {todoStore.loadingTodos}
```
isaachinman: What is happening here? You can't render booleans.

todo-app/src/todo.json
```diff
@@ -0,0 +1,17 @@
+[
```
isaachinman: This file can be removed completely.

todo-app/src/storage.js
```diff
@@ -0,0 +1,8 @@
+export function loadTodoItems() {
```
isaachinman: This file can be removed completely.

PR #50, opened 2018-02-27
add homework

chileluk:


todo-app/src/List-Items.js
```diff
@@ -0,0 +1,30 @@
+import React, { Component } from "react";
+import TodoItems from "./Todos";
+
+export default class ListItems extends Component {
```
isaachinman: What exactly is the purpose of this `ListItems` component? Does it accomplish anything that cannot be accomplished by only having a `Todos` component instead of having both a `ListItems` _and_ a `Todos` component?

Also just a small note, it's convention to have your filename match the component name exactly. Your component name is `ListItems`, so your file should be named `ListItems.js`.

todo-app/src/Todos.js
```diff
@@ -0,0 +1,24 @@
...
+
+export default class TodoItems extends Component {
+
+    render() {
+        const todoItems = this.props.todos;
+        
+        return (
+            <ul>
+                {todoItems.map((item, i) => {
+                    return <li key={i}>
```
isaachinman: Careful about using index as a key - we discussed this in class on week 2 so you probably didn't know yet.

todo-app/src/Todos.js
```diff
@@ -0,0 +1,24 @@
...
+                        {item.task}, {item.deadline}
+                    </li>
+                })}
+            </ul>    
+        );
+    }
+}
+
+
+TodoItems.propTypes = {
```
isaachinman: Nice use of prop validation! If you were using eslint, you might see a warning like:

> todos is not marked as required, and no default value is provided

Or, in other words: "what would happen if this.props.todos was undefined"? You can either set default props or mark this prop as `.isRequired`.

You can check out the [eslint rule here](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/require-default-props.md). This is a little bit more advanced, but something to be aware of!

PR #51, opened 2018-03-02
Submitting week1 homework

talalalamdar:


todo-app/src/App.js
```diff
@@ -1,10 +1,15 @@
 import React, { Component } from 'react';
+import "./style.css"
+import Header from "./components/header.js"
```
isaachinman: These component files should be named the same as the components themselves, please capitalise.

todo-app/src/components/addingContainer.js
```diff
@@ -0,0 +1,43 @@
+
+import React, { Component } from "react";
+import FaPlus from "react-icons/lib/fa/plus"
+
+
+export default class AddingContainer extends Component {
+    constructor(props) {
```
isaachinman: As we discussed in the week 2 class, it's not necessary to use a constructor unless you really want to! Up to you.

todo-app/src/components/addingContainer.js
```diff
@@ -0,0 +1,43 @@
...
+
+    }
+    handleCreating = () => {
+        let descriptionInput = this.state.description
+        let deadLineInput = this.state.deadLine
+         this.props.create(descriptionInput, deadLineInput)
+    }
+
+    render() {
+        let description = <input type="text" value={this.state.description} onChange={this.handleDescriptionChange} />;
```
isaachinman: First, you should be using `const` here instead of `let`, as this value is not being mutated.

Second, it would be more accurate to call these variables something like `descriptionInput` and `deadlineInput`, as you already have `description` and `deadline` as values in your state.

todo-app/src/components/addingContainer.js
```diff
@@ -0,0 +1,43 @@
...
+        this.setState({ description: event.target.value })
+    }
+    handleDeadlineChange = (event) => {
+        this.setState({ deadLine: event.target.value })
+
+    }
+    handleCreating = () => {
+        let descriptionInput = this.state.description
+        let deadLineInput = this.state.deadLine
+         this.props.create(descriptionInput, deadLineInput)
```
isaachinman: Again a small naming issue, please rename `create` to something a little bit more descriptive...

todo-app/src/components/addingContainer.js
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    render() {
+        let description = <input type="text" value={this.state.description} onChange={this.handleDescriptionChange} />;
+        let deadLine = <input type="text" value={this.state.deadLine} onChange={this.handleDeadlineChange} />;
```
isaachinman: So, you're already well beyond week 1's homework! Good job.

At the moment, your deadline input is a text input, and I can type anything into it at all. Here's an example that doesn't make much sense:

screen shot 2018-03-05 at 11 21 32

Let's think about what these deadlines are in the first place - they're dates/date objects, right? I'd like to see you use either a native HTML5 date input, or a third party library (pulled in from `npm`) to allow users to easily select dates from a calendar UI.

todo-app/src/components/todoList.js
```diff
@@ -0,0 +1,45 @@
...
+class TodoList extends Component {
+    constructor(props) {
+        super(props)
+        this.state = {
+            tasks: todoArray
+        }
+    }
+   
+    addingTask = (description, deadLine) => {
+        this.setState((previousState) => ({
```
isaachinman: Careful here. If I'm not mistaken, using this function argument _does not_ enable shallow merging of state. It's also a lot less commonly used.

todo-app/src/components/todoList.js
```diff
@@ -0,0 +1,45 @@
...
+                tasks: [ { description: description, deadLine: deadLine }, ...previousState.tasks]
+            })
+        )
+        console.log(this.state)
```
isaachinman: It doesn't make sense to try and `console.log` out state below your `setState` call, as `setState` is an _asynchronous_ request that does not necessarily execute immediately. A small distinction, but very important to keep in mind.

You should think of `setState` as a future request made to React to update our UI, not an immediate change being made. React will make the requested changes at the next possible moment in terms of performance.

todo-app/src/components/todoList.js
```diff
@@ -0,0 +1,45 @@
...
+    }
+    render() {
+        let tasks = this.state.tasks.map((task, index) =>
+            <TaskItem
+                description={task.description}
+                deadLine={task.deadLine}
+                key={index}
```
isaachinman: As discussed in week 2, don't use index as a key!

talalalamdar:
Thank for the feedback Isaac, i have updated my code, i hope its okay like this, please let me know if you have some more notes

todo-app/src/components/AddingContainer.js
```diff
@@ -22,17 +18,21 @@ export default class AddingContainer extends Component {
     handleCreating = () => {
         let descriptionInput = this.state.description
```
isaachinman: You should be using `const` here, not `let`. I think we've encountered this problem before. You should read [some articles](http://wesbos.com/let-vs-const/) about the difference so you get a good idea why I recommend this.

Also, you could save some time by pulling your variables out as we did in class:

    const { description, deadLine } = this.state

todo-app/src/App.js
```diff
@@ -1,10 +1,68 @@
...
 
 class App extends Component {
+  state = {
+    tasks: myTodoList
+  }
+  componentWillUpdate(nextProps, nextState) {
+    localStorage.setItem("todo list", JSON.stringify(nextState.tasks))
+  }
+  componentDidMount() {
+    if (localStorage.getItem("todo list")) {
```
isaachinman: As shown in class, we can actually do this inside the initial state object. Loading data from localStorage inside `componentDidMount` would hypothetically lead to a flash of different content as the data is loaded in and state is set.

todo-app/src/App.js
```diff
@@ -1,10 +1,68 @@
...
+      this.setState({ tasks: JSON.parse(localStorage.getItem("todo list")) })
+    }
+  }
+  nextId = () => {
```
isaachinman: As discussed in class, this is not a safe way to generate IDs. Please make sure to change this for next week!

todo-app/src/App.js
```diff
@@ -1,10 +1,68 @@
...
+    if (this.state.tasks.length === 0) {
+      return 1;
+    }
+    let uniqueId = this.state.tasks[0].id;
+    return uniqueId += 1;
+  }
+  addingTask = (description, deadLine) => {
+    this.setState((previousState) => ({
+      tasks: [{ id: this.nextId(), description: description, deadLine: deadLine, done: false, creatingDate: moment().format("llll") }, ...previousState.tasks]
```
isaachinman: Would be nice to split this into multiple lines, not one line. It's pretty hard to read this way.

todo-app/src/App.js
```diff
@@ -1,10 +1,68 @@
...
+    })
+    )
+  }
+  handleCheckingBox = (taskId) => {
```
isaachinman: First of all, you are mutating object properties directly instead - you are unintentionally making mutations to your state. Second, `map` exists exactly for this reason, instead of declaring an array and then doing a `forEach`. A better way to write this code would be like:

    handleCheckingBox = (taskId) => {
      const newState = this.state.tasks.map(task => {

        // Returning a _new_ object so that we don't mutate objects inside our state
        return {
          ...task,
          done: taskId === task.id ? !task.done : task.done,
        }
      })

    }

todo-app/src/App.js
```diff
@@ -1,10 +1,68 @@
...
+      (taskId !== task.id) ? task : task.done = !task.done
+    })
+    this.setState({ tasks: newState })
+  }
+  remove = (taskId) => {
+    this.setState(prevState => ({
+      tasks: prevState.tasks.filter(task => task.id !== taskId)
+    }))
+  }
+  edit = (taskId, newDescription) => {
```
isaachinman: Same feedback as for your `handleCheckingBox` function.

todo-app/src/components/AddingContainer.js
```diff
@@ -0,0 +1,47 @@
...
+
+    render() {
+        const descriptionText = <input type="text" value={this.state.description} onChange={this.handleDescriptionChange} />;
+        const deadLineTime = <input type="date" className="deadline-input" onChange={this.handleDeadlineChange} />;
+
+        return (
+            <form onSubmit={this.handleSubmit} className="addingContainer">
+                <label> Description : {descriptionText} </label>
+                <label> Dead line : {deadLineTime} </label>
+                <button type="submit" className="createButton" value="Create new task" onClick={this.handleCreating}>
```
isaachinman: You're doing something weird here... Why do you have _both_ an `onSubmit` and an `onClick` event handler? You should choose one, not both. If you prefer to use the button `onClick` you can simply replace the `form` element with a `div`.

todo-app/src/components/TaskItem.js
```diff
@@ -0,0 +1,66 @@
...
+export default class TaskItem extends Component {
+    static propTypes = {
+        id: PropTypes.number,
+        description: PropTypes.string,
+        deadLine: PropTypes.string,
+        done: PropTypes.bool,
+        creatingDate: PropTypes.string
+    }
+    
+    state = {
```
isaachinman: This state needs to be lifted. I see id, description, deadLine, done, and creatingDate are all coming as props. The `done` value should definitely be a prop as well, with a change function being passed via props too.

isaachinman:
Also, in the future please submit each week's homework as a new PR.

PR #52, opened 2018-03-03
homework react week1

malnajar:


todo-app/build/asset-manifest.json
```diff
@@ -0,0 +1,7 @@
+{
```
isaachinman: So you ran `yarn build` or `npm run build`, causing CRA to build your application for a "production" deployment.

We want to ignore these build files in version control, as they will happen in individual environments depending on the deployment pipeline.

Unfortunately, it looks like the paths in the CRA `.gitignore` are not relative. So, because your `todo-app` project is nested inside another directory, the `.gitignore` rules aren't working correctly.

Can you please correct this, and not commit `build` files in the future? Let me know if you need any help changing the `.gitignore` rules to do this. Also you'll have to exclude previously tracked files via git.

todo-app/package.json
```diff
@@ -1,11 +1,12 @@
 {
-  "name": "todo-app",
+  "name": "week1",
```
isaachinman: Usually the name of our project is going to stay the same. If you're interested in incrementing based on week1, week2, etc, look into incrementing the `version` value using [SEMVER](https://semver.org/).

todo-app/src/App.js
```diff
@@ -1,10 +1,32 @@
...
+import './App.css';
+
+const data = require('./todoList.json');
 
 class App extends Component {
+  state = {
+    todoItems: []
+  }
+  componentDidMount(){
+  this.setState({
```
isaachinman: This is actually unnecessary. It should be possible to simply set your `state` class property above to your imported `data` directly.

todo-app/src/App.js
```diff
@@ -1,10 +1,32 @@
...
+    
     return (
-      <div>
-        Hello World!
+      
+      <div className="App">
+        <Header/>
+        <Container todoItems={this.state.todoItems} />  
+        <Input/>
+        <p className="App-intro">
```
isaachinman: Would be nice to see this removed as it's not part of your todo app.

todo-app/src/Components/Container.js
```diff
@@ -0,0 +1,17 @@
+import React, {Component} from 'react';
+
+export default class Container extends Component {
+
+render() {
+
+    const todos = this.props.todoItems.map((item) => {
+        return <li> {`${item.id}- you have to: ${item.whatToDo} at: ${item.date}`} </li>})
```
isaachinman: As discussed in person, you'll need a `key` prop here.

todo-app/src/Components/Container.js
```diff
@@ -0,0 +1,17 @@
...
+export default class Container extends Component {
+
+render() {
+
+    const todos = this.props.todoItems.map((item) => {
+        return <li> {`${item.id}- you have to: ${item.whatToDo} at: ${item.date}`} </li>})
+    return (
+        <div>
+            <ul>
+            {todos}
```
isaachinman: Watch your indentation!

todo-app/src/Components/Container.js
```diff
@@ -0,0 +1,17 @@
+import React, {Component} from 'react';
+
+export default class Container extends Component {
```
isaachinman: I don't think `Container` is the most accurate name we can come up with here. What does this component do? It renders a `<ul>` of todo `<li>`s. How can we rename this to be more accurate?

todo-app/src/App.js
```diff
@@ -1,10 +1,32 @@
 import React, { Component } from 'react';
+import Header from './Components/Header'
+import footer from './Components/footer'
+import Container from './Components/Container'
+import Input from './Components/Input'
+import './App.css';
+
+const data = require('./todoList.json');
```
isaachinman: So, JavaScript has two main competing module systems. You are free to use either one, but it's best to choose one system and stick to it. In this case, you can replace your CommonJs `require` to something like:

    import data from './todoList.json'

PR #53, opened 2018-03-03
Homework

JawharBairakdar:


documentation/homework/week1/todo-app/src/App.js
```diff
@@ -0,0 +1,22 @@
...
+
+class App extends Component {
+  render() {
+    return (
+      <div className="App">
+        <header className="App-header">
+          <img src={logo} className="App-logo" alt="logo" />
+          <h2>This is Todo-Item App</h2>
+        </header> {/*the Date should be -> M/D/Y */}
+          <Item description='Get out of bed' deadLineDate={'11/13/2017'} isDone={true} />
```
isaachinman: Watch out for indentation here.

documentation/homework/week1/todo-app/src/App.js
```diff
@@ -0,0 +1,22 @@
...
+          <Item description='Brush teeth' deadLineDate={'11/14/2017'} />
```
isaachinman: Hard coding these components for week 1 is fine, but you should start to think about how to render these programmatically inside a loop for the future.

documentation/homework/week1/todo-app/src/components/todo-item.js
```diff
@@ -0,0 +1,24 @@
+import React, { Component } from 'react'
+
+export default class todoItem extends Component {
+  render() {
+    const { description, deadLineDate, isDone} = {
```
isaachinman: I actually have a very hard time understanding what is going on here. If you want to split out variables from `this.props`, usually we do it this way:

    const { description, deadLineDate, isDone } = this.props

If you want to do further modification, like using `toDateString()` etc, you should do that afterwards so it's more clear what's happening.

documentation/homework/week1/todo-app/src/components/todo-item.js
```diff
@@ -0,0 +1,24 @@
...
+      description: this.props.description,
+      deadLineDate: new Date(this.props.deadLineDate).toDateString(),
+      _isDone: this.props.isDone,
+      get isDone() {
```
isaachinman: Again, not really sure what's happening here. I would suggest something like:

    const { isDone } = this.props
    if (isDone()) {
        ...whatever you want to do

documentation/homework/week1/todo-app/src/components/todo-item.js
```diff
@@ -0,0 +1,24 @@
+import React, { Component } from 'react'
+
+export default class todoItem extends Component {
```
isaachinman: React components should always be named with an uppercase first letter. So this component should be called `TodoItem`, and the filename should match: `TodoItem.js`.

documentation/homework/week1/todo-app/src/App.js
JawharBairakdar: I see your point but we don't have Any inserting method to know what kind of data that we will receive

documentation/homework/week1/todo-app/src/components/todo-item.js
JawharBairakdar: yeah sure next time 💔

documentation/homework/week1/todo-app/src/components/todo-item.js
```diff
@@ -0,0 +1,27 @@
+import React, { Component } from 'react'
+
+export default class TodoItem extends Component {
+  render() {
+    const { description, deadLineDate, isDone} = {
+      description: this.props.description,
+      deadLineDate: new Date(this.props.deadLineDate).toDateString(),
+      _isDone: this.props.isDone, // this is a privte property just like state
```
isaachinman: What is your reasoning for setting a prop to a private class property? It's already inside `this.props`. Usually, underscored items are "private" in a different sense - they are internals for libraries which are meant to be ignored by developers etc. Unless you have a very good reason to do this, I would highly recommend you access `this.props.isDone` directly instead of doing this.

PR #54, opened 2018-03-03
week1 react hm

jalalalwani:


isaachinman:
Hi @jalalalwani. How did you go about committing your code in this case? It looks like you only committed/pushed three files. We need a complete CRA application to be able to run your app. We need things like a `package.json`, a `public` directory, etc.

week1_HW/todo-app/src/components/task-list.js
```diff
@@ -0,0 +1,16 @@
+import React, { Component } from 'react'
+import Tasks from "./task"
```
isaachinman: By convention, all React component files should be named exactly the same as the components themselves. So you need to rename these files as `TaskList.js`, `Task.js`, etc.

Also, you're importing the `Task` component as `Tasks` which is not accurate - it only displays a single task.

week1_HW/todo-app/src/components/task-list.js
```diff
@@ -0,0 +1,16 @@
...
+
+export default class TodoList extends Component{
+    render() {
+        return (
+            <ul>
+                <Tasks todo="Attend HYF less" taskTime="2018-03-04" />
```
isaachinman: You have your `todo.json` file, but you aren't using it yet. Instead you're rendering components out statically here. That's fine for week 1, but look into rendering components in a loop for week 2.

week1_HW/todo-app/src/components/task.js
```diff
@@ -0,0 +1,17 @@
+import React, { Component } from 'react'
+
+export default class Tasks extends Component {
```
isaachinman: The component name `Tasks` here is also incorrect, or at least does not match the filename.

isaachinman:
Hi @jalalalwani, I'm going to close this PR as "week1" is in the title. Can you please open new PRs for each week's homework so that we can see what needs to be reviewed?

PR #55, opened 2018-03-05
week 2 HM

yousefallaban:


todo-app/src/index.js
```diff
@@ -1,5 +1,31 @@
-import React from 'react';
+import React, { Component } from 'react';
 import ReactDOM from 'react-dom';
-import App from './App';
+import Todos from './component/todos';
+import Header from "./component/header";
+import JSON from "./todo.json";
```
isaachinman: I'd like to see you name your data differently here. Calling _anything_ `JSON` is going to lead to serious confusion down the line, as `JSON` is a built-in with native methods in JavaScript, eg `JSON.stringify()`. I'm actually pretty surprised you were able to assign a variable to this keyword, never seen this before.

todo-app/src/component/todos.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import TodoItems from './todoItems';
+
+const Todos = ({ todos }) => {
```
isaachinman: Nice use of a functional component!

Just a little trick, you can still do your exporting inline and have an unnamed default export:

    export default ({ todos }) => {

todo-app/src/component/header.js
```diff
@@ -0,0 +1,20 @@
+import React, { Component } from 'react';
+
+class Header extends Component {
```
isaachinman: In some cases you're using functional components, and here you're using a class-based component, even though no state or class properties are used. I understand you may just be experimenting with different ideas, but when you eventually write a "real" codebase, it's a good idea to make a rule for yourself and stick to it.

todo-app/src/component/todoItems.js
```diff
@@ -0,0 +1,44 @@
+import React, { Component } from 'react';
+import Fade from 'react-reveal/Fade';
+
+
+class todoListItem extends Component {
+    state = {
```
isaachinman: This is the only problem I can find in your codebase. Although this technically works, it will cause problems in the future as your app grows.

Why are you storing state in two places? You have your `todos` as state in your `index.js`, which already includes the `done` property, and then you are setting props into state inside your `TodoItem` component and overriding the props.

This is a bad pattern. What if we wanted to show a counter inside `index.js` of how many todos are complete, or how many are incomplete? We would have no way to access the internal state of our children `TodoItem` components.

You need to lift the state from `TodoItem` into `index.js`. Your `handleCheck` event handler should be lifted to `index.js`, and will probably need to take an ID parameter so you can find the appropriate todo. Then, pass this `handleCheck` event handler down your component tree as a prop.

PR #57, opened 2018-03-07
week2 homework

chileluk:


todo-app/src/Add-Todo.js
```diff
@@ -0,0 +1,53 @@
...
+                        key={todo.description}
+                        todoItem={todo} />
+                }
+                )}
+                <div id="form">
+                    <h3>Add new todo item</h3>
+
+                    <textarea value={this.state.description} onChange={this.handleTextChange} />
+
+                    <input type="text" placeholder="MM-DD-YYYY" onChange={this.handleDateChange} />
```
isaachinman: At the moment, your deadline input is a text input, and I can type anything into it at all.

Let's think about what these deadlines are in the first place - they're dates/date objects, right? I'd like to see you use either a native HTML5 date input, or a third party library (pulled in from npm) to allow users to easily select dates from a calendar UI.

todo-app/src/Add-Todo.js
```diff
@@ -0,0 +1,53 @@
...
+        const deadline = e.target.value;
+        this.setState({ deadline });
+    }
+
+    // render new todo when Add button is clicked
+    handleButtonClick = () => {
+        const { description, deadline } = this.state;
+        todoItems.push({ description, deadline, done: false });
+
+        this.setState({ description, deadline });
```
isaachinman: What is this `setState` call? You seem to be fetching the state here, and then setting to state back to what it already was. This is not what we want to do.

In fact, because this is your add-todo function, we want to _reset_ the values here to make the form empty again.

Moreover, in relation to your incorrect state management, this `setState` call is the only reason your app is actually working, as you are storing your `todos` array outside React's state, so it has no idea when this array changes.

todo-app/src/Add-Todo.js
```diff
@@ -0,0 +1,53 @@
+import React from "react";
+import ListItem from "./List-Item";
+
+let todoItems = [];
```
isaachinman: This is wrong. You are keeping track of todos in two separate places: `Main.js` (via `todos.json`) and `Add-Todo.js`, via this array. You basically have two lists of todos.

You need to keep track of your state in a single place. You should import `todos.json` into your `Main.js` component's state, and then add todos there by calling `this.setState`.

todo-app/src/Add-Todo.js
```diff
@@ -0,0 +1,53 @@
...
+
+        this.setState({ description, deadline });
+    }
+
+    render() {
+        return (
+            <section>
+                {todoItems.map(todo => {
+                    return <ListItem
+                        key={todo.description}
```
isaachinman: I'd like to see you use IDs here instead of user-generated input.

todo-app/src/Main.js
```diff
@@ -1,24 +1,52 @@
 import React from 'react';
-import ListItem from "./List-Item";
 import todoItems from "./todos.json";
-import AddTodo from "./Add-Todo";
+import ListItem from "./List-Item";
+import TodoInput from "./Todo-Input";
+
 
+const lastId = todoItems[todoItems.length - 1].id;
```
isaachinman: This is not a safe way to check for IDs. You should use an array method to produce an array of all current IDs, and then use something like `Math.max` to determine the highest value.

If I create four todos from scratch, I'll have todos with IDs of `1, 2, 3, 4`. Now imagine I delete #3, and now have `1, 2, 4`. For the next todo created, your system here would attempt to create another todo with an ID of 4.

todo-app/src/List-Item.js
```diff
@@ -1,6 +1,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 
+
 export default class ListItem extends React.Component {
 
     state = {
```
isaachinman: You're still handling state in two places, keeping the checked state inside the `ListItem`. Is it possible to keep this `checked` state inside the todo objects? What if we needed to list the number of completed todos in `Main.js`? See if you can lift this state up.

PR #58, opened 2018-03-10
Homework react week2

malnajar:


malnajar:
homework week 2

isaachinman:
@malnajar I see several of your commit messages are "Add files via upload", and the others are "Delete {{some component}}". What are you doing? This is not an appropriate way to make commits to a repository. Your commit messages should clearly explain the development work that took place - for example "Added todo creation feature", or "Fixed deadline input bug".

isaachinman:
You have a compilation error that is preventing us from viewing your working app:

screen shot 2018-03-15 at 15 55 40

Please resolve this issue, and comment back here once you've made new commits and pushed them, so that I can review your work.

isaachinman:
Your compilation is still broken. You are attempting to import a file that does not exist (it's named slightly differently). 

Please resolve this issue, and comment back here once you've made new commits and pushed them, so that I can review your work.

todo-app/src/App.js
```diff
@@ -1,10 +1,86 @@
 import React, { Component } from 'react';
+import Header from './Components/Header'
+import Footer from './Components/Footer'
+import Container from './Components/Container'
+
+
+import './App.css';
+const data = require('./todoList.json');
```
isaachinman: Please use one module system as previously discussed.

todo-app/src/App.js
```diff
@@ -1,10 +1,86 @@
...
 
 class App extends Component {
+  state = {
+    todoItems: []
+  }
+  componentWillMount() {
+    localStorage.getItem('todoItems') && this.setState({
```
isaachinman: Please set state directly inside your class property instead of using lifecycle methods.

todo-app/src/App.js
```diff
@@ -1,10 +1,86 @@
...
+      todoItems: JSON.parse(localStorage.getItem('todoItems'))
+    })
+  }
+  componentDidMount() {
+    if (!localStorage.getItem('todoItems')) {
```
isaachinman: Why are you setting state in `componentWillMount` _and_ `componentDidMount`? This is going to cause several renders to happen for no reason. You should initialise your state inside your class property instead of using `setState` in mounting lifecycle methods.

todo-app/src/App.js
```diff
@@ -1,10 +1,86 @@
...
+      this.setState({
+        todoItems: data
+      })
+    } else {
+      console.log('we are using the data from the Local Storage.')
```
isaachinman: You can simply use a `||` operator to set the state as we did in class.

todo-app/src/App.js
```diff
@@ -1,10 +1,86 @@
...
+    }
+
+  }
+  componentWillUpdate(nextProps, nextState) {
+    localStorage.setItem('todoItems', JSON.stringify(nextState.todoItems))
+  }
+
+  handelSelected = (index) => {
+    const todoItems = JSON.parse(localStorage.getItem('todoItems'))
```
isaachinman: This is wrong. You should be trusting React to handle state, not handling it yourself via localStorage. If you want to update your state, you should reference `this.state` to get current values. If you are using localStorage to manage state on your own, you are not using React in the intended way.

todo-app/src/App.js
```diff
@@ -1,10 +1,86 @@
...
+    } else {
+      console.log('we are using the data from the Local Storage.')
+    }
+
+  }
+  componentWillUpdate(nextProps, nextState) {
+    localStorage.setItem('todoItems', JSON.stringify(nextState.todoItems))
+  }
+
+  handelSelected = (index) => {
```
isaachinman: Using index to find an item in an array is a pretty fragile method. Please use an identifier like ID to do this.

todo-app/src/App.js
```diff
@@ -1,10 +1,86 @@
...
+    todoItems.push({
+      "id": this.state.todoItems.length + 1,
+      "description": newTask,
+      "deadline": newDeadLine,
+      "selected": false
+    })
+    this.setState({ todoItems })
+  }
+
+  handelUpdateTask = (index , update) => {
```
isaachinman: Again, please use ID to locate items inside an array, not index. You're already doing this for `handelDeleteTask`, just extend the functionality here.

todo-app/src/App.js
```diff
@@ -1,10 +1,86 @@
...
+
   render() {
+
     return (
-      <div>
-        Hello World!
+
+      <div className="App">
+        <div className="Header"><Header /></div>
+        <div><Container todoItems={this.state.todoItems}
```
isaachinman: It's probably unnecessary to wrap this `Container` component in a div. If you need a div at the top level, just return it at the top level within `Container`.

todo-app/src/App.js
```diff
@@ -1,10 +1,86 @@
...
+    todoItems[index].description = update
+    this.setState({todoItems})
+  }
+
+  handelDeleteTask = (id, selected) => {
+    let todoItems
+    if (selected) {
+      todoItems = this.state.todoItems.filter(task => task.id !== id)
+      todoItems.map((item,index) => {
+        item.id = index + 1
```
isaachinman: This is not a safe way to generate IDs, as we discussed in class.

todo-app/src/App.js
```diff
@@ -1,10 +1,86 @@
...
+    const todoItems = JSON.parse(localStorage.getItem('todoItems'))
+    todoItems[index].selected = !todoItems[index].selected
+    this.setState({ todoItems })
+  }
+
+  handelNewTask = (newTask, newDeadLine) => {
+
+    const todoItems = this.state.todoItems
+    todoItems.push({
+      "id": this.state.todoItems.length + 1,
```
isaachinman: This is not a safe way to generate IDs, as we discussed in class.

todo-app/src/Components/Checkbox.js
```diff
@@ -0,0 +1,22 @@
...
+
+    render() {
+        const {
+            description,
+            selected,
+            index,
+            handelSelected
+        } = this.props
+        return (
+            <div key={index}>
```
isaachinman: Index is not a safe value to use for `key` in JSX as we discussed in class.

todo-app/src/Components/DeleteButton.js
```diff
@@ -0,0 +1,24 @@
+import React, { Component } from 'react';
+
+export default class DeleteButton extends Component {
+    handelDisabledButton = (selected) => {
+        if (selected === false) {
```
isaachinman: You can simply do `return selected === false` instead of these `if else` statements.

PR #59, opened 2018-03-10
homework week2

moesi88:
done


documentation/my-homeworks/week2/todo-app/src/App.js
```diff
@@ -0,0 +1,68 @@
...
+      console.log('we are using the data from the Local Storage.')
+    }
+
+  }
+  componentWillUpdate(nextProps, nextState) {
+    localStorage.setItem('todoItems', JSON.stringify(nextState.todoItems))
+  }
+
+  handelDoneStatus = (index) => {
+    const todoItems = JSON.parse(localStorage.getItem('todoItems'))
```
NouranMahmoud: You should not depend on localStorage at all to get your state, because you already have a state.

First solution
```
const todoItems = this.state.todoItems.map((item, i) => {
      if (i === index) {
        item.done = !item.done;
      }
      return item;
    });
```
Use map to change the required item.

Second solution 
```
const todoItems = [...this.state. todoItems] // clone the array
todoItems[index].done = true
this.setState({todoItems})
```

Third solution

```
const items = this.state.todoItems;

const todoItems = [...items.slice(0, index), 
       {...items[index], done: !items[index]},
        ...items.slice(index + 1)]

this.setState({todoItems});
```

documentation/my-homeworks/week2/todo-app/src/App.js
```diff
@@ -0,0 +1,68 @@
...
+    todoItems[index].done = !todoItems[index].done
+    this.setState({ todoItems })
+  }
+
+  handelNewTask = (newTask) => {
```
NouranMahmoud: Please use meaningful verbs for your functions and avoid using `handle` because it does not reflect the actual action of the function. You can replace it with `addNewTask`

NouranMahmoud: I can see that you are not using it anywhere also.

documentation/my-homeworks/week2/todo-app/src/App.js
```diff
@@ -0,0 +1,68 @@
...
+
+    const todoItems = JSON.parse(localStorage.getItem('todoItems'))
```
NouranMahmoud: As I mentioned above in the previous function, this is not correct to depend on localStorage. Just depend on component state.

documentation/my-homeworks/week2/todo-app/src/App.js
```diff
@@ -0,0 +1,68 @@
...
+    let currentDate = new Date()
+    console.log(currentDate.getFullYear().toString())
+    todoItems.push({
+      "id": this.state.todoItems.length + 1,
```
NouranMahmoud: We agreed about another way to get unique ids in the last session using `Math.max` please use it instead.

documentation/my-homeworks/week2/todo-app/src/App.js
```diff
@@ -0,0 +1,68 @@
...
+  handelDoneStatus = (index) => {
+    const todoItems = JSON.parse(localStorage.getItem('todoItems'))
+    todoItems[index].done = !todoItems[index].done
+    this.setState({ todoItems })
+  }
+
+  handelNewTask = (newTask) => {
+
+    const todoItems = JSON.parse(localStorage.getItem('todoItems'))
+    let currentDate = new Date()
```
NouranMahmoud: why `let` ? use `const` instead if you don't have valid reason for using `let`

documentation/my-homeworks/week2/todo-app/src/components/Container.js
```diff
@@ -0,0 +1,29 @@
...
+            todoItems,
+            handelDoneStatus
+        } = this.props
+
+        return (<div>
+            <ul>
+                {todoItems.map((item, index) => {
+                    return <li key={item.id} className="app-list">
+                        <Checkbox id={item.id}
+                            description={item.description}
```
NouranMahmoud: you can pass all the object once using

`<Checkbox item={item} handelDoneStatus={handelDoneStatus} />`

PR #60, opened 2018-03-11
HW 2

JawharBairakdar:


isaachinman:
Sorry for a late reply on this PR. You have a compilation error that prevents us from viewing your working app in the first place:

screen shot 2018-03-15 at 15 47 35

Simple to fix - please comment back here when you've fixed and pushed new commits, and I'll take a look.

JawharBairakdar:
Hi @isaachinman 
I don't see any conflicting 🙄 [here](https://prnt.sc/irqjwg)
Sorry for delay i am nightly studing :)

JawharBairakdar:
I did a change i hope it will be usefull 🙂

documentation/homework/week1/todo-app/src/components/Todo-item.js
```diff
@@ -0,0 +1,42 @@
+import React, { Component } from 'react'
+
+export default class TodoItem extends Component {
+  constructor(props) {
+    super(props)
+    this.state = {
```
isaachinman: This state probably needs to be lifted to your `App.js` component where you are loading the data from a json file in the first place. I would recommend loading that json file into `App.js`'s state, and then rendering out these `Todo-Item`s with only props and event handlers.

For instance, `App.js` has no way of knowing which todos are checked and which are not. What if you needed to display a total "checked todos" count?

documentation/homework/week1/todo-app/src/components/Todo-item.js
```diff
@@ -0,0 +1,42 @@
...
+    }
+    this.handleCheck = this.handleCheck.bind(this)
+  }
+  handleCheck(e) {
+    const checked = e.target.checked
+    this.setState(prevState => ({
+      checked: !prevState.checked,
+      styleCheck: {
+        textDecoration: checked && 'line-through',
+        color: checked && 'red'
```
isaachinman: Think carefully about this. Why do you need these `styleCheck` properties in your state? Both of them simply respond to the `checked` property, which is already in your state... You are setting two other properties based on `checked` for no reason.

You can update your styling in response to state changes inside the `render` function:

    <li style={this.state.checked ? something : somethingElse}>

Also as a general rule, it's a very bad idea to have CSS declarations inside component state, those two things should not be mixed.

documentation/homework/week1/todo-app/src/App.js
```diff
@@ -0,0 +1,44 @@
+import React, { Component } from 'react'
+import logo from './logo.svg'
+import Item from './components/Todo-item'
+import './App.css'
+import TodoList from './TodosList.json'
+
+
+const Content = () => {
```
isaachinman: Normally it's seen as bad practice to have more than one React component per JavaScript file. You could however put this as a class property of `App` itself, eg:

    renderContent = () => ...your JSX logic

    render() { <div>{this.renderContent()}</div> }

PR #62, opened 2018-03-13
week3 homework

chileluk:


todo-app/src/Main.js
```diff
@@ -0,0 +1,100 @@
...
+        const todoItems = this.state.todos.map(todo => (
+            todo.id === todoId
+                ? { ...todo, done: !todo.done }
+                : todo
+        ));
+        this.setState({ todos: todoItems });
+    }
+
+    addTodo = (description, deadline) => {
+        const todoItems = this.state.todos.map(todo => todo);
```
NouranMahmoud: There's no need for this mapping.

todo-app/src/Main.js
```diff
@@ -0,0 +1,100 @@
...
+        const id = this.state.nextId;
+        this.setState({
+            todos: [
+                ...todoItems,
```
NouranMahmoud: Instead of creating the variables using map, you can do the same like this  `...this.state.todos,`

todo-app/src/Main.js
```diff
@@ -0,0 +1,100 @@
...
+                { id, description, deadline, done: false }
+            ],
+            nextId: id + 1,
+        });
+    }
+
+    toggleEdit = todoId => {
```
NouranMahmoud: No need for this function as I commented before on your `listItem` component.

todo-app/src/Main.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    saveUpdate = (todoId, newDescription, newDeadline) => {
+        const updatedTodos = this.state.todos.map(todo => (
+            todo.id === todoId
+                ? {
+                    ...todo,
+                    description: newDescription,
+                    deadline: newDeadline,
+                    update: !todo.update
```
NouranMahmoud: Don't forget to remove `update` if you will put it as a state inside component.

todo-app/src/Main.js
```diff
@@ -0,0 +1,100 @@
...
+                    }
+                : todo
+        ));
+        this.setState({ todos: updatedTodos });
+    }
+
+    deleteTodo = todoId => {
+        const todoItems = this.state.todos.map(todo => todo);
```
NouranMahmoud: Useless mapping

PR #63, opened 2018-03-14
week3 HW

talalalamdar:


todo-app/src/App.js
```diff
@@ -1,10 +1,85 @@
...
+    )
+  }
+  handleCheckingBox = (taskId) => {
+    const newState = this.state.tasks
+    // eslint-disable-next-line
+    newState.map(task => {
+      // eslint-disable-next-line
+      (taskId !== task.id) ? task : task.done = !task.done
+    })
+    this.setState({ tasks: newState })
```
NouranMahmoud: Here, you use `map` as if it was `forEach`. We use `map` to return new array with mapped values. Like if you have a computation that you need to apply it on all the elements on the array. for example: `[1,2,3,4].map( (i) => i*2);` the return of this will be a new array with `[2, 4, 6, 8]`.

In your code, you don't return anything from the `.map` instead you just mutate/change the state before calling *setState*

Try to add this `console.log` before and after `.map`. You'll find that the second `console.log` prints the `this.state` already changed `done` even before calling `setState`!!

```
handleCheckingBox = (taskId) => {
    const newState = this.state.tasks
    console.log('State Before mutation', this.state.tasks)
    // eslint-disable-next-line
    newState.map(task => {
      // eslint-disable-next-line
      (taskId !== task.id) ? task : task.done = !task.done
    });
    console.log('State After mutation', this.state.tasks);
    this.setState({ tasks: newState })
  }
```

As we agreed we should not update or mutate our state directly without *setState*.

The proper solution is as follows

```
handleCheckingBox = (taskId) => {
    const tasks = this.state.tasks
    console.log('State Before mutation', this.state.tasks)
    // eslint-disable-next-line
    const updatedTasks = tasks.map(task => {
      // eslint-disable-next-line
      if (task.id === taskId) {
        return {
          ...task, done: !task.done
        }
      }

      return task;
    });
    console.log('State After mutation', this.state.tasks);
    this.setState({ tasks: updatedTasks })
  }
```

If you check console, you will find both arrays are the same and the `done` property didn't change at all. But it will change after calling `setState`.

**Please don't hesitate to ask me directly if you still don't understand it.**

todo-app/src/App.js
```diff
@@ -1,10 +1,85 @@
...
+  }
+  remove = (taskId) => {
```
NouranMahmoud: Leave new line between each function please.

todo-app/src/App.js
```diff
@@ -1,10 +1,85 @@
...
+    const newState = this.state.tasks.filter(task => task.id !== taskId)
+    this.setState({
+      tasks: newState
+    })
+  }
+  edit = (taskId, newDescription) => {
```
NouranMahmoud: This's the same as `handleCheckingBox` function. Please reconsider it.

todo-app/src/App.js
```diff
@@ -1,10 +1,85 @@
...
+    const newState = this.state.tasks
+    // eslint-disable-next-line
+    newState.map(task => {
+      // eslint-disable-next-line
+      (taskId !== task.id) ? task : task.description = newDescription
+    })
+    this.setState({ tasks: newState })
+  }
+  handleDeletingAllCompleted = () => {
```
NouranMahmoud: Functions naming should be meaningful verbs to what the function does. `handle` does not represent what the function want to do, So please give naming more care because this is a good sign of a professional programmer ;)

You can name it `deleteAllCompleted`.

todo-app/src/components/AddingContainer.js
```diff
@@ -0,0 +1,49 @@
...
+        const deadLineInput = this.state.deadLine
+        if (descriptionInput && deadLineInput) {
+            this.props.onCreate(descriptionInput, deadLineInput)
+        } else {
+            alert("you should fill the values of description and deadline")
+        }    
+    }
+    handleSubmit = (e) => {
+        e.preventDefault()
+        this.handleCreating()
```
NouranMahmoud: There's no need here to create new function `handleCreating`. You can just add the logic of creating here.

todo-app/src/components/DeleteButton.js
```diff
@@ -0,0 +1,16 @@
+import React, { Component } from "react";
+
+export default class DeleteButtom extends Component {
+    onDeletingAll = () => {
+        this.props.onDeletingAll()
+    }
+    render() {
+        return (
+            <div className="delete-container">
+                <button className="delete-all-button" onClick={ this.onDeletingAll }>
```
NouranMahmoud: IF there's no specific logic of your event in this component, there will be no need for creating new function. You can just use `this.props.onDeleteAll()` directly like the following.

```
<button className="delete-all-button" onClick={ this.props.onDeletingAll }>
```

PR #65, opened 2018-03-16
adding hm3 reactjs

jalalalwani:
sorry for being late


isaachinman:
There is no `package.json` in your `week3_HM` directory, so no one can run this project at all. Please fix.

week3_HM/todo/src/components/TodoList.js
```diff
@@ -0,0 +1,79 @@
+import React, { Component } from 'react'
+import input from './buttons'
+import InsertTasks from './insertTasks'
+//import DatePicker from 'react-datepicker';
+//import moment from 'moment';
+
+const data = require('../todo.json')
```
isaachinman: Please use one module system, not both. You can easily accomplish the same thing here with:

    import data from '../todo.json'

week3_HM/todo/src/components/TodoList.js
```diff
@@ -0,0 +1,79 @@
...
+
+export default class TodoList extends Component{
+    state = {
+     mineTasks: data
+    }
+    handleChecked = (index) => {
```
isaachinman: Index is not a very safe way to be looking up items in an array. Please refactor to use ID.

week3_HM/todo/src/components/TodoList.js
```diff
@@ -0,0 +1,79 @@
...
+        this.setState({
+            mineTasks
+       })
+    }
+    handleAddingNewTask = (text, date) => {
+        console.log(text, date)
+        const ids = this.state.mineTasks.map(task => task.id)
+        const newId = Math.max(...ids) + 1
+        const newTask = {
+            "id": newId,
```
isaachinman: No need for quotes on properties in JavaScript objects unless they contain characters like `-`.

week3_HM/todo/src/components/TodoList.js
```diff
@@ -0,0 +1,79 @@
...
+            "deadline": date,
+            "done": false
+        }
+        this.setState({
+            mineTasks: [
+                ...this.state.mineTasks,
+                newTask]
+        })
+    }
+    handleDeleteTask = (index) => {
```
isaachinman: Again, using index to look up items in an array is not safe. The best way to perform deletion in this case is to pass ID as the argument to your function, and then use a simple `array.filter` and set state.

week3_HM/todo/src/components/Task.js
```diff
@@ -0,0 +1,23 @@
+import React, { Component } from 'react'
```
isaachinman: It looks like you're no longer using this component in week 3? If so, please remove it from your codebase.

week3_HM/todo/src/components/TodoList.js
```diff
@@ -0,0 +1,79 @@
...
+            "description": text,
+            "deadline": date,
+            "done": false
+        }
+        this.setState({
+            mineTasks: [
+                ...this.state.mineTasks,
+                newTask]
+        })
+    }
```
isaachinman: Please be careful with indentation!

week3_HM/todo/src/components/TodoList.js
```diff
@@ -0,0 +1,79 @@
...
+    handleDeleteTask = (index) => {
+        console.log("deleting")
+        const mineTasks = this.state.mineTasks
+        const newMIneTasks = mineTasks.splice(index  , 1)
+        this.setState({
+        newMIneTasks
+    })
```
isaachinman: Another indentation issue. Take a look at eslint, it will handle these sorts of problems for you by producing warnings.

week3_HM/todo/src/components/TodoList.js
```diff
@@ -0,0 +1,79 @@
+import React, { Component } from 'react'
+import input from './buttons'
+import InsertTasks from './insertTasks'
```
isaachinman: As we discussed in class, js files for React components should match the component name. In this case you need to capitalise it.

week3_HM/todo/src/components/insertTasks.js
```diff
@@ -0,0 +1,48 @@
...
+
+
+export default class InsertTasks extends Component {
+
+    state = {
+        taskValue: '',
+        dateValue: ''
+    }
+
+    handleTextChange = (event) => {
```
isaachinman: Look at your `handleTextChange` function and your `handleDateFieldChange` function. Besides naming, they do exactly the same thing.

Have you heard of DRY as a programming principle? It means "don't repeat yourself". It's not a huge problem here, but you might consider in the future how to combine these two functions into a single function - imagine you had 10 different inputs, not just 2.

PR #69, opened 2018-03-17
HW 3

JawharBairakdar:
Hi @isaachinman 
I am sorry i didn't modify the Change that you request i found it hard to figuering out and i have to practice more so i just did it again.


documentation/homework/week3/todo/src/App.js
```diff
@@ -0,0 +1,84 @@
...
+// Components
+import AssignItem from './components/Assign'
+import ListItem from './components/ListItem'
+
+// Params
+import Params from './Params/locals'
+
+class App extends Component {
+  state = {
+    list: []
```
isaachinman: Can you use a more accurate/descriptive name for this array?

documentation/homework/week3/todo/src/App.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+  componentWillMount() {
+    this.setState({
+      list: Params.getLocals || []
+    })
+  }
+  Moments(date) {
```
isaachinman: Not totally sure what is going on here, but I think you would solve this problem by simply using a standard date input instead of a text input.

documentation/homework/week3/todo/src/components/Assign.js
```diff
@@ -0,0 +1,20 @@
...
+
+class AssignItem extends Component {
+
+  render() {
+    //x -- 2x inputs
+    //x -- 1x button
+    return (
+      <form onSubmit={(event) => this.props.handleAssign(event)}>
+        <InputField name="description" />
+        <InputField name="deadline" />
```
isaachinman: So you're using a regular text input here for a date field. It's not great, as I can type "blahblahblah" and then click add new todo, and I end up seeing UI that says "Invalid date". Please look into [HTML5 date inputs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date), or even check out 3rd party React datepicker components on npm.

documentation/homework/week3/todo/src/App.js
```diff
@@ -0,0 +1,84 @@
...
+  handleRemove = (event, local_id) => {
+    this.setState({
+      list: Params.removeLocal(local_id)
+    })
+  }
+  handleAssign = (event) => { // Inserting Values
+    event.preventDefault()
+    const { deadline, description } = event.target
+    this.setState({
+      list: Params.saveLocal( // the Id is automaticly generats
```
isaachinman: This is a pretty strange way to structure your updates. Inside `saveLocal`, you're actually writing data to `localStorage`, and then getting that data _back_ from localStorage and returning it. I would not recommend this, you should just set your React state directly, and then _respond_ to state changes inside lifecycle methods like `componentDidUpdate`.

In other words, you should just let React do its own thing. LocalStorage is only a "nice" thing we can add on top, it shouldn't be core functionality responsible for reading and writing state.

I would suggest you rewatch the video from Week 3, specifically how we implemented localStorage there. You have a bunch of localStorage functions like `removeLocal`, `updateLocal`, etc etc. These are all totally unnecessary and not good practice. You should be writing these functions inside your React component as handlers, not externally.

Then, you can just respond on each update cycle and write the state to localStorage, no matter what it is.

documentation/homework/week3/todo/src/Params/locals.js
```diff
@@ -0,0 +1,54 @@
+
+class Params {
+  get random_8() { // just for Id's For Now
```
isaachinman: This is probably a "safe" way to generate unique IDs, without looking into it too deep... But just so you are aware, there are lots and lots of popular libraries to do things like this, where another developer has spent many hours solving annoying problems so that you don't have to.

I think [uuid](https://www.npmjs.com/package/uuid) is probably the most popular. 1.5 million downloads per day.

documentation/homework/week3/todo/src/Params/locals.js
```diff
@@ -0,0 +1,54 @@
...
+    }
+    return randomString
+  } // this function should return a random string that 8 Characters long
+
+  // localStorage section
+  get getLocals() {
+    const container = localStorage.getItem('list')
+    return JSON.parse(container)
+  }
+  saveLocal(getLocals, local_Object) {
```
isaachinman: Why are you passing a class property back into a class property?

documentation/homework/week3/todo/src/App.js
```diff
@@ -0,0 +1,84 @@
...
+        Params.getLocals,
+        {
+          description: description.value,
```
isaachinman: In your creation function here, it's important to reset the form values after we create a new todo, so that the form is reset and the inputs are empty. Right now the values stay inside the input, so if I want to add a second new todo, I have to delete my old values first.

documentation/homework/week3/todo/src/components/ListItem.js
```diff
@@ -0,0 +1,66 @@
...
+class ListItem extends Component {
+  // The Worst practice Ever
+  state = {
+    edit: false
+  }
+  
+  handleSubmit = (event) => {
+    const { id } = this.props
+    event.preventDefault()
+    this.props.handleUpdateListItem(event, id, 'description', event.target.description.value)
```
isaachinman: You are not using controlled inputs! This is not the React way! You are accessing values as form data onSubmit instead of passing a `value` and a `onChange` prop to your input.

Additionally, because of this, when I click "edit" on a todo item, I see empty text fields instead of the existing values which should be there.

Please make these inputs controlled.

documentation/homework/week3/todo/src/App.js
JawharBairakdar: Yes sure I will 👍

documentation/homework/week3/todo/src/Params/locals.js
JawharBairakdar: it's just the parameter name, But you are right i should name it something else because it's still confusing. 👍

isaachinman: No... if you take a look here:

https://github.com/JawharBairakdar/React/blob/007658ae5b1a44cf26dd6182eed6f2eb64fd9b2d/documentation/homework/week3/todo/src/App.js#L35

It looks like you're actually passing the class property back in as an argument. Unless I am missing something which is of course possible!

JawharBairakdar:
I tried to fix what you have mentioned to in this Homework 🙂

isaachinman:
Hey @JawharBairakdar can you open an new PR to reflect Week 4 HW?

PR #70, opened 2018-03-17
week3

naghamfarwati:


documentation/week3/src/components/Form.js
```diff
@@ -0,0 +1,31 @@
...
+
+    render() {
+        return (
+            <div className='Form'>
+                <form onSubmit={(event) => this.props.handleSubmit(event)}>
+                    
+                    <label> Task :    
+                       <input type='text' placeholder='discription' value={this.props.discription}
+                            onChange={(event) => this.props.handleDiscriptionChange(event)} 
+                        discription={this.props.discription}
```
NouranMahmoud: Based on our previous comment for `changeState(key, value)`. You can use it here like this `this.props.changeState('discription', event.target.value)}

documentation/week3/src/App.js
```diff
@@ -0,0 +1,83 @@
...
+
+  handleDiscriptionChange = (event) => {
+    event.preventDefault();
+    this.setState({
+      discription: event.target.value
+    })
+  }
+
+  handleDeadlineChange = (event) => {
+    event.preventDefault();
```
NouranMahmoud: There's no need for `event.preventDefault()` here. We use it only when there's default behaviour for the HTML element that we are using, like `<form></form>`, because it has default behaviour.

Before Ajax, we were using <form method="GET" action="/submitAction.php"> tag with <input type="submit"> to make Backend calls directly without writing javascript code. That's why we're using `event.preventDefault()` .. to prevent this default behaviour.

documentation/week3/src/components/Form.js
```diff
@@ -0,0 +1,31 @@
+import React, { Component } from 'react';
+
+export default class Form extends Component {
+
+    render() {
```
NouranMahmoud: Use this `const {discription, deadline} = this.props;` instead of writing `this.props.whatever` multiple times. **This is better performance wise and also code readability**

documentation/week3/src/components/List.js
```diff
@@ -0,0 +1,30 @@
+import React, { Component } from 'react';
+import Task from './Task';
+
+export default class List extends Component {
```
NouranMahmoud: Better naming can be `Tasks` or `TasksList`

documentation/week3/src/components/List.js
```diff
@@ -0,0 +1,30 @@
...
+   
+    render() {
+        
+        return (
+            <div className='List'>
```
NouranMahmoud: Don't use capital letters `List` with classes. only small letters, but if you want to add many words just combine them with `-` or `_`

documentation/week3/src/components/List.js
```diff
@@ -0,0 +1,30 @@
...
+                
+                {this.props.todos.map((todo, id) => {
+                    
+                    return (
+                <Task
+                    key={id}
+                    id ={id}
```
NouranMahmoud: I think you wanted to use `todo.id` here.

documentation/week3/src/App.js
```diff
@@ -0,0 +1,83 @@
...
+    this.setState({
+      deadline: event.target.value
+    })
+  }
+
+  handleClick = (id) => {
+    const todos = this.state.todos;
+    todos[id].done = !todos[id].done;
```
NouranMahmoud: By doing this you are mutating your state without using `this.setState`. We avoid mutating state because it will produce un-expected bugs also React will use its power of managing state using `this.setState`.

To understand this just write the following `console.log`:-

**Bad example**
```
const todos = this.state.todos;
    todos[id].done = !todos[id].done;
    console.log(todos[id]); // true
    console.log(this.state.todos[id]); // true
```
By doing this you will find that `this.state.todos[id]` already changed. However it shouldn't change until we use `this.setState`.

**Good example**
```
    const todos = this.state.todos.map( (todo) => (todo.id === id ? todo : {...todo, done: !todo.done}))
    console.log(todos[id]); // false
    console.log(this.state.todos[id]); // true
```
By deleting your code and just using maps, you will find that `this.state.todos[id]` didn't change and it will change only when you use `this.setState`.

documentation/week3/src/App.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  handleSubmit = (event) => {
+    event.preventDefault();
+    const newTodo = {
+      discription: this.state.discription,
+      deadline:this.state.deadline,
+      done: false
+    };
+    const todos = this.state.todos;
```
NouranMahmoud: The same as the previous comment here, `const todos = this.state.todos;` this doesn't mean you created a **clone** from the todos object but it means you just created another **reference** for it. So when you change the variable **todos**, it will change the original **this.state.todos** because your new variable is just a **reference** .

So that, `todos.push(newTodo)` will push this `newTodo` to the **this.state.todos**.

To make a clone from an array you can do like this `const todos = [...this.state.todos];`

documentation/week3/src/App.js
```diff
@@ -0,0 +1,83 @@
...
+  render() {
+
+    return (
+
+      <div className="App">
+
+        <Form
+          handleDiscriptionChange={this.handleDiscriptionChange}  
+          handleDeadlineChange={this.handleDeadlineChange}
+          inputValue={this.state.inputValue}
```
NouranMahmoud: What's this **inputValue**, I think you meant to send `discription` and `deadline` here ;)

PR #71, opened 2018-03-18
new todo app

moesi88:


mkruijt:
hi @moesi88 why did you close your pr?

moesi88:
Cuz when i push iT at the class i cheacked iT with Issac and it was empty and I dnt knw why -__- 
I should push it now when I get home to let Issac and Noran let some feedbacks on it ...

PR #72, opened 2018-03-20
Week4 HW

yousefallaban:
handel all the state by Mobx stores and mutate it  withe MobX actions


todo-app/src/App.js
```diff
@@ -0,0 +1,23 @@
...
+import { Provider } from 'mobx-react'
+
+// Components
+import Header from "./component/Header";
+import Todos from "./component/Todos";
+
+export default class App extends React.Component {
+  render() {
+    return (
+      <div>
```
isaachinman: Why are you wrapping everything in a div? There should be no need for that.

todo-app/src/stores/TodoStore.js
```diff
@@ -0,0 +1,54 @@
+import { action, observable } from 'mobx';
+import moment from 'moment';
+
+// Data
+import TodoData from '../data/todo.json';
+
+export class TodoStore {
+  @observable todos = TodoData
+  @observable newDescription = ''
+  @observable newDeadline = moment().format('YYYY-MM-DD')
```
isaachinman: You're setting the default todo deadline to today. Are you sure this is what you want to do? Why not leave it blank?

todo-app/src/stores/TodoStore.js
```diff
@@ -0,0 +1,54 @@
...
+
+  @action
+  handelDeleteItem = (todo) => {
+    this.todos = this.todos.filter((item) => {
+      return item.id !== todo.id
+    })
+  }
+
+  @action
+  handleDescriptionChange = (event) => {
```
isaachinman: It looks like your `handleDescriptionChange` and `handleDateChange` functions can be abstracted into a single function. Remember DRY - don't repeat yourself.

todo-app/src/stores/TodoStore.js
```diff
@@ -0,0 +1,54 @@
...
+
+  @action
+  handleDateChange = (event) => {
+    this.newDeadline = event.target.value
+  }
+
+  // add new todo
+  @action
+  handleSubmit = (event) => {
+    event.preventDefault();
```
isaachinman: In my opinion, using `event.preventDefault` in a MobX action is incorrect. This is of course somewhat an opinion, but MobX actions should be totally independent of the UI. They should be (ideally) pure functions that take an input and mutate state based on that input.

If you want to use a form and call `preventDefault`, it is probably better to do this outside your MobX action, and just pass the value in here.

It actually makes me realise we probably shouldn't be passing `event` objects into these MobX actions in the first place, and should only pass the actual values. I will try to remember to discuss this in class.

todo-app/src/stores/TodoStore.js
```diff
@@ -0,0 +1,54 @@
...
+    const ids = this.todos.map((todo) => todo.id);
+    const newId = this.todos.length > 0 ? Math.max(...ids) + 1 : 1;
+    const newTodo = {
+      id: newId,
+      description: this.newDescription,
+      deadline: this.newDeadline,
+      done: false
+    };
+    this.todos = [newTodo, ...this.todos];
+    this.newDescription = '';
```
isaachinman: Would be nice to see the default values for `newDescription` and `newDeadline` stored somewhere instead of hardcoding them in multiple places.

todo-app/src/component/Todos.js
```diff
@@ -0,0 +1,50 @@
...
+      newDescription,
+      todos,
+    } = this.props.TodoStore;
+
+    const hasTodos = todos.length > 0;
+    
+    return (
+      <div className="todosContainer">
+        <h3>Todo-List</h3>
+        <AddTodoForm description={newDescription}
```
isaachinman: So you are connecting your `Todos` component to the store, but then passing a lot of things as props instead of connecting the child components directly. Both options work, but I wonder if you had a reason for this?

The main point of using a `Provider` is that we can connect child components anywhere in the tree to the data they need, even if it's just one field from the store.

todo-app/src/component/TodosItem.js
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+
+//Component
+import EditForm from './EditForm';
+import DisplayForm from './DisplayForm';
+
+
+export default class TodosItem extends React.Component {
```
isaachinman: A few points to make about this component.

First of all, you decided to put the "new todo" form state inside your MobX store as `newDescription` and `newDeadline`. Why did you decide to keep the editing values here as local React component state? You should choose one approach to form state in your application and stick to it instead of this inconsistency.

Second, you need to initialise all your state values inside your `state =` class property. You're setting state values here that aren't declared initially, and that leads to confusion for other developers and your future self.

Third, there is no reason to be setting state based on props in `componentDidMount`. You can do this directly inside your state class property.

Fourth, it appears that you are not passing the changed/saved values of your todo items into your store _at all_. You are only saving them in component local state. Take a look at your `handelSubmit` function. While it is sometimes acceptable to keep local state in form components, as soon as the data is "submitted" or "finalised", it needs to be integrated into the store.

PR #73, opened 2018-03-21
week 4 homework

chileluk:


todo-app/src/stores/TodosStore.js
```diff
@@ -0,0 +1,70 @@
+import todos from "../data/todos.json";
+import { observable, action, configure } from "mobx";
+
+configure({ enforceActions: true });
+
+
+const TodoItems = observable({
```
isaachinman: While this does "work", it's not really the correct way to be working with MobX without decorators. Check out [this section of the docs](https://mobx.js.org/best/decorators.html).

Instead of just wrapping your `todos` array in `observable`, you're wrapping the entire object. Moreover, it's customary to use classes, not plain objects.

todo-app/src/stores/TodosStore.js
```diff
@@ -0,0 +1,70 @@
...
+    },
+
+    addTodo: action((id, description, deadline) => {
+        if (description && deadline) {
+            const newTodo = {
+                id,
+                description,
+                deadline,
+                done: false, 
+                editMode: false
```
isaachinman: I see you've got for the "multiple" editing way instead of the single editing way. There's a small problem with embedding this kind of editing logic directly in our data - it will have to be added and removed each time we interface with an API. You'll have to do some extra refactoring next week.

todo-app/src/stores/TodosStore.js
```diff
@@ -0,0 +1,70 @@
...
+            }
+            TodoItems.todos.push(newTodo);
+        }
+    }),
+
+    toggleCheck: action(todoID => {
```
isaachinman: Looks to me like your `toggleCheck` and `toggleEdit` could be abstracted into a single function, eg `toggleProperty`.

todo-app/src/stores/TodosStore.js
```diff
@@ -0,0 +1,70 @@
...
+    }),
+
+    saveUpdate: action((todoID, description, deadline) => {
+        if (description && deadline) {
+            TodoItems.todos = TodoItems.todos.map(todo => (
+                todo.id === todoID
+                    ? { ...todo, description, deadline }
+                    : todo
+            ));
+            TodoItems.toggleEdit(todoID);
```
isaachinman: Why don't you just manually set the editing to false? Calling "toggle" seems to be introducing the possibility of bugs here.

todo-app/src/stores/TodosStore.js
```diff
@@ -0,0 +1,70 @@
...
+        }   
+    }),
+
+    deleteTodo: action(todoID => {
+        TodoItems.todos = TodoItems.todos.filter(todo => (
+            todo.id !== todoID
+        ));
+    }),
+
+    get doneCount() {
```
isaachinman: These getters should be wrapped/decorated with [computed](https://mobx.js.org/refguide/computed-decorator.html) so that MobX can optimise your functions.

todo-app/src/index.js
```diff
@@ -1,5 +1,7 @@
-import React from 'react';
-import ReactDOM from 'react-dom';
-import App from './App';
+import React from "react";
+import ReactDOM from "react-dom";
+import "./index.css";
+import App from "./components/App";
```
isaachinman: Normally it's customary to keep your React root (eg App.js) in the root directory, and put sub-components into a components directory. This way we know immediately who is the parent.

todo-app/src/components/ListItem.js
```diff
@@ -0,0 +1,64 @@
...
+
+    const date = deadline.split("-").join("/");
+    const crossOut = done ? "crossOut" : null;
+
+    return (
+        <section>
+            {
+                editMode
+                ?
+                <UpdateTodo todo={todo}
```
isaachinman: You're passing redundant data here, no? You can either pass the individual todo object and the update function, or the entire store and just the todo's ID.

todo-app/src/components/UpdateTodo.js
```diff
@@ -0,0 +1,44 @@
+import React from "react";
+
+
+export default class UpdateTodo extends React.Component {
+
+    state = {
+        description: this.props.todo.description,
```
isaachinman: What made you decide to keep `editMode` in your MobX store, but keep the actual editing data in local state? Could you defend this decision in a code review?

todo-app/src/index.js
```diff
@@ -1,7 +1,14 @@
...
+import { Provider } from "mobx-react";
+import todoItems from "./stores/TodosStore";
+import App from "./App";
 import "./index.css";
-import App from "./components/App";
-import TodoItems from "./stores/TodosStore";
 
-ReactDOM.render(<App todoItems={TodoItems}/>, document.getElementById("root"));
+const TodoApp = () => (
+    <Provider todoItems={todoItems}>
```
isaachinman: It's more appropriate to call this `todos` or `todoStore` rather than `todoItems`, as this store contains more than just the items themselves, but also actions, etc.

todo-app/src/components/TodoList.js
```diff
@@ -0,0 +1,36 @@
...
+import { inject, observer } from "mobx-react";
+import ListItem from "./ListItem";
+
+
+@inject("todoItems")
+@observer    
+export default class TodoList extends React.Component {
+
+    render() {
+        const { todoItems } = this.props;
```
isaachinman: You are destructuring twice for no reason. You can just do:

    const { todos, doneCount } = this.props.todoItems

todo-app/src/components/TodoList.js
```diff
@@ -0,0 +1,36 @@
...
+        const { todos, doneCount } = todoItems;
+        
+        return (
+            <section>
+                { todos.length
```
isaachinman: Brackets shouldn't be padded with spaces.

todo-app/src/components/ListItem.js
```diff
@@ -1,64 +1,62 @@
...
+            editTodo,
+            deleteTodo,
+        } = this.props.todoItems;
+
+        const {
+            id,
+            description,
+            deadline,
+            done
+        } = this.props.todo;
```
isaachinman: You're passing the whole `todo` object, but also injecting the store, so you have duplication of data here. If you want to inject the store into this component, it would be better to only pass `id` as a prop into `ListItem`, and lookup the item in the store inside `ListItem` itself.

todo-app/src/components/ListItem.js
```diff
@@ -1,64 +1,62 @@
...
+
+        const date = deadline.split("-").join("/");
+        const crossOut = done ? "crossOut" : null;
+
+        return (
+            <section>
+                { editID === id
+                    ?
+                    <UpdateTodo todo={this.props.todo} />
```
isaachinman: Remember that you can spread objects into props too, to avoid having to do `this.props.todo.id`. You can do:

    <UpdateTodo {...this.props.todo} />

Just an option!

todo-app/src/components/AddTodo.js
```diff
@@ -1,38 +1,39 @@
...
                 />
                 <input type="date"
                     value={deadline}
-                    onChange={(e) => this.changeInput(e, "deadline")}
+                    onChange={(e) => {
+                        changeAddFormInput(e.target.value, "deadline")
+                    }}
                 />
                 <button onClick={() => {
                     addTodo(nextId, description, deadline)
```
isaachinman: Because all these values are available inside your store already, there's no need to pass them as arguments into the `addTodo` function. Just access them directly inside your store.

PR #74, opened 2018-03-22
new todo week3 hm sorry for dalay bt i had many problem with pushing my files bt this is my  app tht i showed u in class last sunday ...

moesi88:


isaachinman:
There are two folders, `week-3` and `week3.2`. Can you delete the incorrect one and update the PR?

documentation/my-homeworks/week3.2/src/components/app.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import CreateTodo from './create-todo';
```
NouranMahmoud: It's better to name *Components* as names not verbs, like `TodoForm` not `CreateTodo`

documentation/my-homeworks/week3.2/src/components/app.js
```diff
@@ -0,0 +1,50 @@
...
+import TodosList from './todos-list';
+import todos from './todos.json'
+
+
+export default class App extends React.Component {
+    state = {
+            myTodos:todos
```
NouranMahmoud: You named your state key as `myTodos` but you are using it as `this.state.todos` in most of your code.

PR #75, opened 2018-03-23
HW 4

JawharBairakdar:


documentation/homework/week4/src/App.js
```diff
@@ -0,0 +1,25 @@
...
+
+// Import Stores Index
+import Stores from './stores'
+
+// Import Components Index
+import Index from './components'
+
+class App extends Component {
+  render() {
+    // Sorry... I don't have an OOP background Am I that good?! :D
```
isaachinman: Haha!

documentation/homework/week4/src/App.js
```diff
@@ -0,0 +1,25 @@
...
+import './App.css'
+
+// Import Provider
+import { Provider } from 'mobx-react'
+
+// Import Stores Index
+import Stores from './stores'
+
+// Import Components Index
+import Index from './components'
```
isaachinman: You should call this something more descriptive. A component should never be called Index.

documentation/homework/week4/src/stores/InitialStore.js
```diff
@@ -0,0 +1,102 @@
...
+// set the default state of the items Object
+const default_item_state = {
+  id: 0,
+  title: '',
+  deadline: '',
+  done: false,
+  Edit: false,
+}
+
+class Counter {
```
isaachinman: First, you forgot to rename the class when copy pasting! Second, why are you calling this store `InitialStore`?

documentation/homework/week4/src/components/index.js
```diff
@@ -0,0 +1,43 @@
...
+import ViewRender from './ViewRender'
+import EditRender from './EditRender'
+import { locals } from '../utils'
+
+@inject('InitialStore')
+@observer
+export default class extends Component {
+  componentDidUpdate = (nextProps) => locals.save = [...nextProps.InitialStore.items]
+  render() {
+    const { items, removeItem, toggle_edit, toggle_checkbox, submit_edit, item_edit_state, onInputEdit } = this.props.InitialStore
```
isaachinman: So you are connecting your `Index` component to the store, but then passing a lot of things as props instead of connecting the child components directly. Both options work, but I wonder if you had a reason for this?

The main point of using a Provider is that we can connect child components anywhere in the tree to the data they need, even if it's just one field from the store.

documentation/homework/week4/src/components/index.js
```diff
@@ -0,0 +1,43 @@
...
+import { inject, observer } from 'mobx-react'
+
+import AssignItem from './AssignItem'
+import ViewRender from './ViewRender'
+import EditRender from './EditRender'
+import { locals } from '../utils'
+
+@inject('InitialStore')
+@observer
+export default class extends Component {
```
isaachinman: You should give this class a name.

documentation/homework/week4/src/components/index.js
```diff
@@ -0,0 +1,43 @@
...
+  render() {
+    const { items, removeItem, toggle_edit, toggle_checkbox, submit_edit, item_edit_state, onInputEdit } = this.props.InitialStore
+    return (
+      <div>
+        <AssignItem />
+        <ul>
+          {items.map(item => {
+            return (
+              (item.Edit) ?
+                <EditRender
```
isaachinman: There's no need to include "render" in your React component names... All components render things!

documentation/homework/week4/src/components/index.js
```diff
@@ -0,0 +1,43 @@
...
+
+import AssignItem from './AssignItem'
+import ViewRender from './ViewRender'
+import EditRender from './EditRender'
+import { locals } from '../utils'
+
+@inject('InitialStore')
+@observer
+export default class extends Component {
+  componentDidUpdate = (nextProps) => locals.save = [...nextProps.InitialStore.items]
```
isaachinman: This works, but I'd rather see you use MobX's [autorun](https://mobx.js.org/refguide/autorun.html) function to achieve the same thing.

documentation/homework/week4/src/stores/InitialStore.js
```diff
@@ -0,0 +1,102 @@
...
+    const newItem = {
+      ...this.item_state,
+      id: locals.id
+    }
+    this.items = [
+      ...this.items,
+      newItem,
+    ]
+    this.item_state = default_item_state
+    return this.items
```
isaachinman: MobX actions don't need to return anything. Any particular reason you are doing this?

documentation/homework/week4/src/stores/InitialStore.js
```diff
@@ -0,0 +1,102 @@
...
+  onInputChange = (e, field) => (this.item_state = {
+    ...this.item_state,
+    [field]: e.target.value
+  })
+
+  @action
+  addNewItem = () => {
+    const newItem = {
+      ...this.item_state,
+      id: locals.id
```
isaachinman: Why is your id generation logic inside your localstorage utils? Should be moved somewhere else, probably inside the store here.

documentation/homework/week4/src/stores/InitialStore.js
JawharBairakdar: Well First i didn't copy paste! and i was just following the Counter lesson making things simple.
second I couldn't find a name 🙃

documentation/homework/week4/src/components/index.js
JawharBairakdar: Actully my reason that I hate the code that has too many imports inside it and some times I feel it very messy, Well you might say we have the `@inject` and the `@observer` too.
But It's just a one level and there is not that many `props` I am passing for each one.

And i don't know i just felt more comfortable working with the child component, because the hall almost work in two places the stores and the last child component.

PR #76, opened 2018-03-23
Week 4's HW

malnajar:


isaachinman:
Hi @malnajar, is your homework supposed to be complete, or are you still working on it?

I just checked out this PR, and saw local state still present inside `App.js`. I see you've added a `TasksListStore` and have connected one component, but the MobX migration does not seem complete.

todo-app/src/Stores/index.js
```diff
@@ -0,0 +1,9 @@
+import { configure } from 'mobx'
+
+import TasksListStore from './TasksListStore'
+
+configure({ enforceActions: true })
+
+export default {
+    data: TasksListStore,
```
isaachinman: You shouldn't be calling your store `data`... Imagine we had a few other stores. You can name the key here exactly the same as the store if you like, or use camel case. Up to you.

todo-app/src/Stores/TasksListStore.js
```diff
@@ -0,0 +1,151 @@
...
+    editedTask = ''
+
+    @observable
+    newTask = {
+        description: '',
+        date: '',
+    }
+
+    @action
+    selected = (id) => {
```
isaachinman: The naming here could be a bit better. It's _toggling_ the selection of todo items.

todo-app/src/Stores/TasksListStore.js
```diff
@@ -0,0 +1,151 @@
...
+        const todoItems = this.todoList.map(task => {
+            if (task.id === id) {
+                return {
+                    "id": task.id,
```
isaachinman: No need for quotes on these.

todo-app/src/Stores/TasksListStore.js
```diff
@@ -0,0 +1,151 @@
...
+                    "selected": !task.selected,
+                    "isDone": task.isDone,
+                    "timeout": task.timeout
+                }
+            } else {
+                return task
+            }
+        })
+        this.todoList = todoItems
+        saveToLocalStorage(this.todoList)
```
isaachinman: Instead of having to put `saveToLocalStorage` at the end of each of your actions, you could use `autorun` to automatically save the store to localStorage each time it changes.

todo-app/src/Stores/TasksListStore.js
```diff
@@ -0,0 +1,151 @@
...
+        const todoItems = this.todoList
+        const id = this.todoList.map(task => task.id)
+        const newId = this.todoList.length > 0 ? Math.max(...id) + 1 : 1
+        todoItems.push({
+            "id": newId,
+            "description": newTask.toUpperCase(),
+            "deadline": newDeadLine,
+            "selected": false,
+            "isDone": false,
+            "timeout": this.setTimeout(newDeadLine)
```
isaachinman: You should definitely choose a new name for `this.setTimeout`. `setTimeout` is already a built-in method to JavaScript, and does something very very different from your class method here.

todo-app/src/Stores/TasksListStore.js
```diff
@@ -0,0 +1,151 @@
...
+
+    @action
+    setNewTask = (event, field) => {
+        this.newTask[field] = event.target.value
+    }
+
+    @action
+    setTimeout = (deadline) => {
+        let taskDeadline = moment(deadline).toObject()
+        if (moment([taskDeadline.years, taskDeadline.months, taskDeadline.date]).diff(moment(), 'days') < 0) {
```
isaachinman: Unnecessary use of an `if/else`. You can just do:

    return moment([taskDeadline.years, taskDeadline.months, taskDeadline.date]).diff(moment(), 'days') < 0

Also, you can definitely do this without having to use `toObject()`. Here's a quick example:

    moment(deadline).diff(moment(), 'days')

todo-app/src/Stores/TasksListStore.js
```diff
@@ -0,0 +1,151 @@
...
+        saveToLocalStorage(this.todoList)
+    }
+
+    @action
+    setNewTask = (event, field) => {
+        this.newTask[field] = event.target.value
+    }
+
+    @action
+    setTimeout = (deadline) => {
```
isaachinman: You should probably just do this calculation in your UI. You're unnecessarily storing two values relating to the same thing: `deadline` and `timeout`. As long as we have `deadline`, we can always determine whether or not the deadline is the past pretty easily... We don't need two fields in our store.

todo-app/src/Stores/TasksListStore.js
```diff
@@ -0,0 +1,151 @@
...
+
+        const todoItems = this.todoList.filter(task => task.id !== id)
+        this.todoList = todoItems
+        saveToLocalStorage(this.todoList)
+    }
+
+    @action
+    selectToEdit = (id) => {
+        this.selectedToEdit = id
+        this.editedTask = ''
```
isaachinman: It would be nice to preserve the existing value of the todo when I click edit, so that I don't have to start from scratch every time. We did this with the blog app in class, can you recreate that functionality here?

todo-app/src/Stores/TasksListStore.js
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    @action
+    updateTask = (id, update) => {
+        const todoItems = this.todoList.map(task => {
+            if (task.id === id) {
+                return {
+                    "id": task.id,
+                    "description": update,
```
isaachinman: You seem to be using these full objects each time, even when updating a single field. You should use either the spread operator, or `Object.assign`. Here's an example of spreading:

    return {
      ...task,
      description: update,
    }

PR #77, opened 2018-03-24
week4 homework

talalalamdar:


todo-app/src/components/AddingContainer.js
```diff
@@ -0,0 +1,34 @@
...
+    handleSubmit = (e) => {
+        e.preventDefault()
+        this.props.todosStore.createTodo();
+    }
+
+    render() {
+        const { createTaskForm, editingDescription, editingDeadline } = this.props.todosStore
+        const { description } = createTaskForm 
+
+        const descriptionText = <input placeholder="enter a description" onChange={e => editingDescription(e)} />;
```
NouranMahmoud: If you are going to send **e** only as a parameter, you can do it like this `nChange={editingDescription}`

todo-app/src/components/TaskItem.js
```diff
@@ -0,0 +1,78 @@
...
+
+    static propTypes = {
+        id: PropTypes.number,
+        description: PropTypes.string,
+        deadLine: PropTypes.string,
+        done: PropTypes.bool,
+        creatingDate: PropTypes.string
+    }
+
+    handleToggling = () => {
```
NouranMahmoud: Naming function should start with meaningful verbs for the real action that the function does. For example here it should be `toggleTodo`. For removing you can name it `removeTodo`. In general don't use **handle** because it is vague for what your function really does.

todo-app/src/stores/TodosStore.js
```diff
@@ -0,0 +1,167 @@
...
+
+    @action
+    onChangeEditingText = (e) => {
+        this.editingText = e.target.value
+    }
+
+    @action
+    enableEditMode = (id) => {
+        this.taskEditingId = this.todosList.filter(task => task.id === id)[0].id
+        this.editingText = this.todosList.filter(task => task.id === id)[0].description
```
NouranMahmoud: You can use the function `this.todoList.find` instead of `this.todosList.filter` because find will not return array like filter. ;)

todo-app/src/components/FeaturesContainer.js
```diff
@@ -0,0 +1,30 @@
...
+import { inject, observer } from "mobx-react"
+
+@inject("todosStore")
+@observer
+export default class FeaturesContainer extends Component {
+
+    render() {
+        const { todosList, deleteAllCompleted, markAllTodos, unMarkAllTodos } = this.props.todosStore
+
+        const completedTasks = todosList.filter(task => task.done === true).length
```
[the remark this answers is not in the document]
talalalamdar: thanks for the feedback Nouran, they were helpful, i had updated my code according to that, let me know if you have more notes

PR #78, opened 2018-03-24
Adding HM week4.

jalalalwani:


week2_HM/src/App.js
```diff
@@ -0,0 +1,19 @@
...
+        <TodoList/>
+      </div>
+    );
+  }
+}
+
+export default App;
+
+
+//dataFromApp = { this.state.isDone }
```
isaachinman: Please remove things like this before opening a PR.

week4_HW/mytodo-app/src/components/TodoList.js
```diff
@@ -0,0 +1,46 @@
+import React, { Component } from 'react'
+import input from './buttons'
+import InsertTasks from './insertTasks'
+import {inject, observer} from 'mobx-react'
+
+@inject("data")
+@observer
+export default class TodoList extends Component{
+    render() {
+        console.log(this.props.data)
```
isaachinman: Please remove console logging before you submit your work!

week4_HW/mytodo-app/src/components/TodoList.js
```diff
@@ -0,0 +1,46 @@
...
+        const { deletingTask, checkingBox, addingNewTask} = this.props.data
+        const myTasks = this.props.data.tasksData.map((task, index) => {
+            if (task.done === true) {
+                return (
+                    <li key={task.id}
+                        className="unDoneTask"
+                        style={{ "textDecoration": "line-through" }}>
+                        <input type="checkbox"
```
isaachinman: This approach is broken. You have an uncontrolled input here with an onChange handler. You are not controlling the checked value of the input directly. The only reason this works is because you're using `defaultChecked`, but this is not a safe approach. If you are setting the "done" status of your todos to `false` automatically, you already have everything you need to create a proper controlled input.

It seems like you are switching over `task.done` and rendering big pieces of JSX based on this checked value. Instead I'd like to see conditional rendering inside your JSX, something like:

    <li key={task.id}
        className={task.done ? "" : "unDoneTask"}
    >
        <input type="checkbox"
            onChange={() => checkingBox(index)}
            checked={task.done} />
    {index+1}- {task.description} on {task.deadline}
        <input type="button" value="Remove"
            onClick={() => deletingTask(index)} />
    </li>

week4_HW/mytodo-app/src/stores/index.js
```diff
@@ -0,0 +1,13 @@
...
+import TaskList from './TaskList'
+import { configure } from 'mobx'
+
+
+configure({ enforceActions: true })
+
+export default  {
+
+    values: DifferentValues,
+    data: TaskList
```
isaachinman: These `values` and `data` names are not appropriate names for stores. All stores presumably have values and data in them. You should name your stores based on the category of data they are storing.

week4_HW/mytodo-app/src/stores/DifferentValues.js
```diff
@@ -0,0 +1,27 @@
+
+import {observable, action} from 'mobx'
+
+class DifferentValues {
```
isaachinman: Why did you create a separate store for this? If it works, that's OK, I am just wondering your reasoning behind this.

week4_HW/mytodo-app/src/stores/TaskList.js
```diff
@@ -0,0 +1,44 @@
...
+    @action
+    checkingBox = (index) => {
+        const mineTasks = this.tasksData
+        mineTasks[index].done = !mineTasks[index].done;
+        return mineTasks
+    }
+    @action
+    addingNewTask = (text, date) => { 
+        const newTask = {
+            "id": moment().toString(),
```
isaachinman: No need for quotes on object keys.

PR #79, opened 2018-03-28
week5 homework

talalalamdar:


todo-app/src/stores/TodosStore.js
```diff
@@ -0,0 +1,172 @@
+import { observable, action, runInAction } from "mobx"
+
+
+
```
isaachinman: No more than 2 empty lines in a row please!

todo-app/src/stores/TodosStore.js
```diff
@@ -0,0 +1,172 @@
...
+                body: JSON.stringify(this.createTaskForm)
+            })
+            runInAction(() => {
+                // reset the form
+                this.createTaskForm = createTaskFormDefault
+            })
+        } else {
+            alert("you should fill the values of Description and Deadline")
+        }
+        this.getTodos()
```
isaachinman: I think your `this.getTodos()` call should be inside your `if` block, not here at the end. You're making API calls even when the user hasn't input the correct fields yet.

todo-app/src/stores/TodosStore.js
```diff
@@ -0,0 +1,172 @@
...
+                'content-type': "application/json"
+            },
+            body: JSON.stringify({ done: !task.done })
+        })
+        this.getTodos()
+
+    }
+
+    @action
+    deleteAllCompleted =  () => {
```
isaachinman: Looks like this action is not making an API calls. Is this still pending? A nice way to approach this would be to use `Promise.all` and make as many concurrent requests as necessary to make the action as fast as possible. It would look something like:

    await Promise.all(todosToDelete.map(todo => this.deleteTodo(todo.id)))

Assuming the `deleteTodo` function returns a `Promise`, of course!

todo-app/src/stores/TodosStore.js
```diff
@@ -0,0 +1,172 @@
...
+        const newTodosList = this.todosList.filter(task => task.done === false)
+        this.todosList = newTodosList
+
+
+    }
+
+    @action
+    addingDescription = value => {
```
isaachinman: It looks like your `addingDescription` and `addingDeadline` functions can be abstracted into a single function.

todo-app/src/stores/TodosStore.js
```diff
@@ -0,0 +1,172 @@
...
+        this.createTaskForm.description = value
+    }
+
+    @action
+    addingDeadline = value => {
+        this.createTaskForm.deadline = value
+    }
+
+    @action
+    markAllTodos = () => {
```
isaachinman: Also not connected to the API, you could use the same approach as batch deletion.

talalalamdar:
thanks for the feedback, i have implement some of your tips, but i didn't know how to do api calls using promise.all , so i used forEach, i hope it's okay like this

todo-app/src/stores/TodosStore.js
```diff
@@ -0,0 +1,171 @@
...
+        })
+        this.getTodos()
+
+    }
+
+    @action
+    deleteAllCompleted = async  () => {
+        const completedTodos = this.todosList.filter(task => task.done === true)
+         await completedTodos.forEach(task => {
+             fetch(`${API_ROOT}/todos/${task._id}`, {
```
isaachinman: You are not actually awaiting anything here, as `forEach` does not return a promise. Moreover, you are calling `getTodos` with _each_ `/todos/${task._id}` call on line 86, but not even waiting for the initial API call on line 83 either!

Lots of bugs here, but programming concurrent network requests is not beginner material for sure.

You essentially have two general options with this:

1. Use `await` in a loop (eg for loop or while loop) and process network requests sequentially. Pros: more hands-on control of error handling, also good if requests depend on each other. Cons: slow!

2. Use `Promise.all` for concurrency. Pros: enables multiple network requests at once - browsers were made for this. Cons: `Promise.all` is fail-fast, meaning if one request fails, any that have not yet been sent are cancelled (there are ways around this).

This is how Option 2 would look:

    await Promise.all(completedTodos.map((async todo =>
      await fetch(`${API_ROOT}/todos/${task._id}`, {
        method: "DELETE"
      })
    )))

    await this.getTodos()

Notice how I am directly returning an `await` call from the map. This is the same as returning a promise. Actually, the `await` is implied in a one-liner like this, but that's separate. Remember that `async` functions return promises. Also note how we only call `getTodos` at _the end_, after all our requests are complete. This code sample assumes your `getTodos` method returns a promise.

Here's how Option 1 would look:

    for (let i = 0; i < completedTodos.length; i++) {
      await fetch(`${API_ROOT}/todos/${completedTodos[i]._id}`, {
        method: "DELETE"
      })
    }

    await this.getTodos()

Yes, a good old fashioned for loop. I would advise you to _not write code like this_, though. A lot of senior developers might get mad if they saw something like this. The reason a for loop works, while `forEach` does not has to do with scope. Since `forEach` calls a _new_ function for each item, that new function is not an `async` function. Of course you can make it `async`, but then you have nested awaiting and weird stuff. A for loop will work fine, but most people will tell you awaiting inside a loop is something you should never do.

A lot of this has to do with more complicated topics on concurrency and the event loop, so don't worry too much if it seems crazy right now.

talalalamdar: yeah i agree, i noticed when i wrote it, it really looks weird, about nested awaiting, but i dont understand why you said it is not working, it is making api calls and changing the data in the api, i tried first to await on fetch but then i got an error that await is reserved word, but apparently i had to make the callback async, i think it was a problem with scope

and i actually read about about promise.all and i understood the concept and i saw simple examples but i couldnt find a real code example about api calls, so i gave up, but your example above is understood.

Thanks for the feedback Isaac, it was really helpful :) . i will implement it right away

isaachinman: > but i dont understand why you said it is not working, it is making api calls and changing the data in the api

Yes, it's "working" but it is very bug prone. The two main problems currently:

1. No good way to handle errors
2. Calling `this.getTodos()` for _every_ item in the loop

PR #80, opened 2018-03-28
Week 5 Homework MobX Async Actions 

yousefallaban:


todo-app/src/stores/TodoStore.js
```diff
@@ -0,0 +1,123 @@
...
+  todoEditingID = null
+
+  @observable
+  todoEditingText = ''
+
+  //TODOs API
+  @action
+  getTodos = async () => {
+    const res = await fetch(`${API_ROOT}/todos`);
+    const parseResponse = await res.json()
```
isaachinman: Please watch out for your formatting. On line 26 you use semicolons, and on this line you don't. Both methods are acceptable because we are in a transpiled environment, but definitely choose one and stick to it!

todo-app/src/stores/TodoStore.js
```diff
@@ -0,0 +1,123 @@
...
+    this.todos = this.todos.map((task) => {
+      if (task._id === todo._id) {
+        return {
+          ...task,
+          done: !task.done
+        }
+      }
+      return task
+    })
+    await fetch(`${API_ROOT}/todos/${todo._id}`, {
```
isaachinman: What happens if this network request fails? You need add a `catch` and unset the checkbox change made in this case, otherwise you will have incorrect data being displayed to the user.

todo-app/src/stores/TodoStore.js
```diff
@@ -0,0 +1,123 @@
...
+        'content-type': 'application/json'
+      },
+      body: JSON.stringify({
+        done: !todo.done
+      })
+    })
+  }
+
+  @action
+  DeleteTodoItem = async (todoId) => {
```
isaachinman: Actions should be `camelCase`.

todo-app/src/stores/TodoStore.js
```diff
@@ -0,0 +1,123 @@
...
+      method: 'PATCH',
+      headers: {
+        'content-type': 'application/json'
+      },
+      body: JSON.stringify({
+        description: this.todoEditingText
+      })
+    })
+    this.getTodos()
+    this.cancelEditing()
```
isaachinman: In this case, we're not really "cancelling" the editing, we're "done" editing. So maybe just rename this action to `closeEditing` or `endEditing` or something.

todo-app/src/stores/TodoStore.js
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  // editing todo
+
+  @action
+  toggleEdit = (id) => {
+    this.todoEditingID = id
+    this.todoEditingText = this.todos.filter(todo => {
+      return todo._id === id
+    })[0].description
```
isaachinman: Instead of doing `myArray.filter(() => {})[0]`, you can use `myArray.find`. Check out the [docs here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find).

PR #82, opened 2018-04-04
HW 5

JawharBairakdar:


documentation/homework/week4/src/stores/todo_actions.js
```diff
@@ -0,0 +1,150 @@
...
+// Setting our localStorage
+// import { locals } from '../utils'
+
+// set the default state of the items Object
+const default_item_state = {
+  description: '',
+  deadline: '',
+}
+
+async function request(url, method, inputToString) {
```
isaachinman: This is a really nice abstraction. But I'd rather see it in your `utils` directory, not inside the store. Imagine we had several stores, each one would want to use `request`.

documentation/homework/week4/src/stores/todo_actions.js
```diff
@@ -0,0 +1,150 @@
...
+    const data = await request('todos')
+    runInAction(() => {
+      this.items = data
+    })
+  }
+
+  @action // onInput Change
+  onInputChange = (value, field) => {
+    this.item_state = {
+      ...this.item_state,
```
isaachinman: Why set the entire object? Why not use square bracket syntax? Eg:

    this.item_state[field] = value

documentation/homework/week4/src/stores/todo_actions.js
```diff
@@ -0,0 +1,150 @@
...
+  switch (method) {
+    case "POST":
+    case "PATCH":
+      options = { method, ...params, }
+      break
+    default:
+      options = { method }
+  }
+
+  // is't Okay to catch the errors here with in
```
isaachinman: Not entirely sure what you are asking but I'll respond in general:

Errors "bubble". If you don't catch them inside your `request` helper, you'll need to catch them one level up, eg wherever you call `request()`. Either way is fine, but typically you'd expect a helper like `request` to pass errors up through invocation, not handle them itself. Imagine we were using `request` in multiple stores, and each store wanted to handle errors in a slightly different way.

The only hard rule is that you _must_ catch errors somewhere. Uncaught rejections will soon be a breaking runtime error in Node.

documentation/homework/week4/src/stores/todo_actions.js
```diff
@@ -0,0 +1,150 @@
...
+
+// set the default state of the items Object
+const default_item_state = {
+  description: '',
+  deadline: '',
+}
+
+async function request(url, method, inputToString) {
+  const params = {
+    "body": JSON.stringify(inputToString),
```
isaachinman: You don't need to put quotes around object keys!

documentation/homework/week4/src/stores/todo_actions.js
```diff
@@ -0,0 +1,150 @@
...
+      description,
+      deadline,
+    }
+    this.items = this.items.map(item => { // Rendering the applied changes
+      if (item._id === _id) {
+        return {
+          ...item,
+          ...applyChange,
+          // no need for existing (Edit) in the original schema any way
+          Edit: false,
```
isaachinman: Key names should be lowercase unless they're `SPECIAL_CONSTANTS`.

documentation/homework/week4/src/stores/todo_actions.js
```diff
@@ -0,0 +1,150 @@
...
+  // is't Okay to catch the errors here with in 
+  // -- one single line with the.json line ?
+  // OR we have to catch it on the fetch it self also ?!
+  // Because what i see here is a strange look to me
+  const preResponse = await fetch(`https://hyf-react-api.herokuapp.com/${url}`, options)
+  const response = await preResponse.json().catch(console.error)
+  return response
+}
+
+class todo_actions {
```
isaachinman: This is not a good name for your store. Your store contains a lot more than just actions! It also contains observables, and can contain computed values etc. It should be called `TodoStore` in capitalised case like that.

documentation/homework/week4/src/stores/todo_actions.js
JawharBairakdar: thank you! i didin't notice that 😅

documentation/homework/week4/src/stores/todo_actions.js
JawharBairakdar: yeah thank you Isaac I got it 🙂

documentation/homework/week4/src/stores/todo_actions.js
JawharBairakdar: this is just a styling thing because i didn't like the style that some with an other without like the following
```
const params = {
      body: JSON.stringify(inputToString),
      headers: {
        "Content-type": "application/json"
      }
    }
```

JawharBairakdar:
Actulley I am so lazy to change the name of the file i had wokin' on so i just kept it as it's and also if i changed the name of the folder it will appear like a new folder in github.

That proves for me working every day and questions are not useless 😃 
Thank you! Happy to hair that 🙂

PR #84, opened 2018-04-06
new todo-apps

moesi88:
I did fix my apps problems and build a new bulletin app that fetch data from sever and today ill try to do it with mobx and load the modifications to my repo

so kindly if that possible to just check it and advice me with some notes 

thanks a lot


documentation/my-homeworks/bulletin-board/src/Board.js
```diff
@@ -0,0 +1,86 @@
...
+		this.nextId = this.nextId.bind(this)
+	}
+
+	componentWillMount() {
+		var self = this
+		if(this.props.count) {
+			fetch(`https://baconipsum.com/api/?type=all-meat&sentences=${this.props.count}`)
+				.then(response => response.json())
+				.then(json => json[0]
+								.split('. ')
```
NouranMahmoud: This long white space usually happens because you are using *tabs* instead of *spaces* on your editor, so please stick to *spaces*.

documentation/my-homeworks/bulletin-board/src/Note.js
```diff
@@ -0,0 +1,100 @@
...
+			transform: `rotate(${this.randomBetween(-25, 25, 'deg')})`
+		}
+	}
+
+	randomBetween(x, y, s) {
+		return x + Math.ceil(Math.random() * (y-x)) + s
+	}
+
+	componentDidUpdate() {
+		var textArea
```
NouranMahmoud: Don't use *var*, also there's no need to define the variable before the If condition because you use the variable only inside its block.

documentation/my-homeworks/bulletin-board/src/Note.js
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	edit() {
+		this.setState({
+			editing: true
+		})
+	}
+
+	remove() {
+		this.props.onRemove(this.props.index)
```
NouranMahmoud: If you are going to make a call `this.props.onRemove(this.props.index)` only, so there's no need to create **remove** function. Just use the `this.props.onRemove(this.props.index)` directly on JSX code.

moesi88:
Tnx a lot Nour cuz u did that i had problem with {npm run Eject } bt now its working ill mention tht i upload my changes when am done 

And thts true I should not use var anymore and really am using Tab to much when am typing I’ll be worry about tht in next upload 


Tnx again and hav a gd night

PR #86, opened 2018-04-07
week4

naghamfarwati:


naghamfarwati:
Hi @isaachinman yes,, but i have to fix the button of do it toggleDone action ,,after i fix the delete button, the Done button didn’t work,, can you help me to fix it??

isaachinman:
@naghamfarwati Your problem lies in the fact that you copy/pasted code from our in-class work to your homework.

I would suggest you start with ctrl+f in your project, searching for "isLiked".

naghamfarwati:
@isaachinman yes, but that after i tried with my function alot... and when it didn’t work, i tried to do the same function that you did in the class,,, any way today i will send another solution to you 😊

isaachinman:
The problem is a simple naming problem, you are calling things "done" and "isLiked" at the same time. Just sort out your naming and you'll find that it already works!

PR #88, opened 2018-04-16
Empty app skeleton for todo app

n4cr:
Empty project created by create-react-app for handing out the homework


isaachinman:
Hey @n4cr and @mkruijt - why was this skeleton added to the documentation dir? First, students should get used to using CRA via the command line by themselves, and second it's causing a lot of confusion as to where students' homework should be located inside the repo.

I think instead we should have a top-level dir called `homework-submission` or something like that. Anything inside that dir we can assume is homework, and students shouldn't be submitting PRs on any code outside that dir.

n4cr:
hey @isaachinman there was a dead link in the [documentation ](https://github.com/HackYourFuture/React/blob/master/documentation/homework/W1-homework.md)for week 1 homework to a todo app regarding how to deliver the homework:
```
    Fork this repository and base your week 1 homework on the skeleton todo-app.
```
That's why I added this as I didn't want to change the structure. Feel free to update it if it's making it confusing for the students.

isaachinman:
Okay, got it!

mkruijt:
Hi @isaachinman , what Nasir said! I think it's a good idea to make a dir specifically for homework hand in. To avoid confusion in the future. Will you update the repo or do you prefer me to do it?

isaachinman:
@mkruijt Either you or I can do it, but it's too late with class 13. Adding a new dir, etc, would cause confusion - the students would have to rebase their forks. Let's do it at the end of this module?

mkruijt:
yes makes sense, would be great if you can do it. Another thing database teachers are now altering their assignments so that they add on/ use the code in the Node module. Would be cool if we can do the same for React once database is updated. Than the students have a working  have a full-stack web app. Also there are a lot of comments in the homework about proper formatting, in Node and JavaScript we solve this by adding a standard VSCode folder with auto formatting and a `.eslintrc` file. For example here: https://github.com/HackYourFuture/JavaScript2/blob/master/.eslintrc . Maybe we can add this to the react Module as well? Let me know what you think.

remarcmij:
These are the VSCode **workspace** settings that I'm proposing to include in all PR-based homework repos:

```json
{
  "editor.formatOnSave": true,
  "editor.formatOnType": true,
  "editor.formatOnPaste": true,
  "editor.detectIndentation": false,
  "editor.tabSize": 2
}
```

Since all students now use VSCode, this should ensure consistent formatting, even if the global VSCode settings are missing or differ.

mkruijt:
To answer your question, students are already introduced to `eslintrc` in the javascript modules (and Node), so that should not be an issue. We've just set up travis for our new website, never worked with CircleCI though. But I agree is a good plan. And I also agree that this discussion is getting of topic. I will create some issues, let's further discuss this in the mentor room in slack

PR #91, opened 2018-04-19
homework week1 hamza zydan

zydan1:


documentation/todo-app/src/App.js
```diff
@@ -1,19 +1,37 @@
 import React, { Component } from 'react';
-import logo from './logo.svg';
 import './App.css';
-
+import Todo from './Todo';
 class App extends Component {
+  
   render() {
+    let tasks = [
```
isaachinman: A couple things here:

1. You should be using `const`, not `let`, as this array is not mutated.

2. You should declare this variable _outside_ your render function. As you have it right now, the variable is _redeclared_ each time the render function runs. This is a small issue, but would have performance issues in some cases.

documentation/todo-app/src/App.js
```diff
@@ -1,19 +1,37 @@
...
-          <img src={logo} className="App-logo" alt="logo" />
-          <h1 className="App-title">Welcome to React</h1>
-        </header>
-        <p className="App-intro">
-          To get started, edit <code>src/App.js</code> and save to reload.
-        </p>
-      </div>
+     <div className = "root">
+       {tasks.map((ele,i)=>
+<Todo task = {ele.task} date = {ele.date} deadline = {ele.deadline} key= {i}/>
```
isaachinman: Be very careful of using index as a key in React. It is much better to give your tasks a unique ID, and use that as their key.

Also, please be careful of your indentation and formatting!

documentation/todo-app/src/App.js
```diff
@@ -1,19 +1,37 @@
...
 import './App.css';
-
+import Todo from './Todo';
 class App extends Component {
+  
   render() {
+    let tasks = [
+      {
+        task: " Get out of bed",
+        date: " Wed Sep 13 2017",
```
isaachinman: Interesting choice to have both `date` and `deadline`. Do you mean a creation timestamp, and the due date for the task to be completed? You might want to update the names to make this more clear.

documentation/todo-app/src/Todo.js
```diff
@@ -0,0 +1,22 @@
+import React , {Component} from 'react';
+
+
+export default class Todo extends Component{
+  
+render(){
+ let {task,date,deadline} = this.props;
```
isaachinman: Again, you should be using `const`. As a general rule, you should _always_ use `const` unless you are forced to use `let`.

documentation/todo-app/src/Todo.js
```diff
@@ -0,0 +1,22 @@
+import React , {Component} from 'react';
+
+
+export default class Todo extends Component{
+  
+render(){
+ let {task,date,deadline} = this.props;   
+ return(
+<div className = "todo">
```
isaachinman: The formatting in this file is quite bad and needs improvement.

documentation/todo-app/src/Todo.js
```diff
@@ -0,0 +1,22 @@
...
+<h3> {task}</h3>
+<span> Date : <strong>{date}</strong></span>
+<h4>Deadline: {deadline}</h4>
+
+</div>
+
+ )   
+}
+
```
isaachinman: Please, no more than 2 blank lines in a row.

PR #98, opened 2018-04-30
Elmira _react HW1 (HW2 , HW3 added here)

elmira202003:
This is the homework week 1 , and I am so sorry for delay!
thanks in advance for your checking.

Week2 & week3 HWs have committed to this request.

### DIFF
--- a/documentation/todo-app/src/components/todoItems.js
+++ b/documentation/todo-app/src/components/todoItems.js
@@ -1,0 +1,22 @@
+import React, { Component } from 'react';
+
+export default class TodoItems extends Component {
+    render() {
+        const { description, deadline } = this.props;
+        
+        return (
+            <div> 
+                <tbody>
+                    <tr>
+                        <td className = "description">
+                           { description }
+                        </td>
+                        <td className = "deadline">
+                           { deadline }
+                        </td>
+                    </tr>
+                </tbody>  
+            </div>
+        )
+    }
+}
--- a/documentation/todo-app/src/components/todoMain.js
+++ b/documentation/todo-app/src/components/todoMain.js
@@ -1,0 +1,33 @@
+import React, { Component } from 'react';
+import TodoItems from './todoItems';
+import './todo_main.css'
+
+export default class TodoMain extends Component {
+    render() {
+        const todos = [
+            {
+                'description': 'Get out of bed',
+                'deadline': 'Wed Sep 13 2017'
+            },
+            {
+                'description': 'Brush teeth',
+                'deadline': 'Thu Sep 14 2017'
+            },
+            {
+                'description': 'Eat breakfast',
+                'deadline': 'Fri Sep 15 2017'
+            }
+        ]
+
+        const main_todos = todos.map((todo, i) => (
+            <TodoItems description = { todo.description } deadline = { todo.deadline } />
+        )
+        );
+
+        return (
+            <div>
+                 { main_todos }
+            </div> 
+        )
+    } 
+}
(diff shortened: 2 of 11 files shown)


documentation/todo-app/src/components/todo_main.css
```diff
@@ -0,0 +1,34 @@
+
```
isaachinman: Your css normally should not be inside your `components` directory, unless it's structured in a different way. Css is not a component, so it shouldn't be here!

documentation/todo-app/src/components/todoMain.js
```diff
@@ -0,0 +1,33 @@
+import React, { Component } from 'react';
```
isaachinman: Your React components should have a filename which is (exactly) the same as the component name. We always capitalise React component names, so it's also conventional to capitalise the filename. In this case, you should have:

    Header.js => not headers.js
    Logo.js => not logo.js
    TodoItem.js => not todoItems.js (it is a singular item, not a list)
    TodoMain.js => not todoMain.js

documentation/todo-app/src/components/todoMain.js
```diff
@@ -0,0 +1,33 @@
...
+import TodoItems from './todoItems';
+import './todo_main.css'
+
+export default class TodoMain extends Component {
+    render() {
+        const todos = [
```
isaachinman: You should declare this variable outside your render function. As you have it right now, the variable is redeclared each time the render function runs. This is a small issue, but would have performance issues in some cases.

documentation/todo-app/src/components/todoMain.js
```diff
@@ -0,0 +1,33 @@
+import React, { Component } from 'react';
+import TodoItems from './todoItems';
+import './todo_main.css'
```
isaachinman: Please be careful with your semicolon use. You can choose to either use semicolons or not use semicolons, but whichever way you choose, you must be consistent.

documentation/todo-app/src/components/todoMain.js
```diff
@@ -0,0 +1,33 @@
...
+                'deadline': 'Thu Sep 14 2017'
+            },
+            {
+                'description': 'Eat breakfast',
+                'deadline': 'Fri Sep 15 2017'
+            }
+        ]
+
+        const main_todos = todos.map((todo, i) => (
+            <TodoItems description = { todo.description } deadline = { todo.deadline } />
```
isaachinman: First, we don't normally put spaces around the `=`. Second, if you run your app and then open the dev console in whatever browser you are using, you'll see this warning:

> Warning: Each child in an array or iterator should have a unique "key" prop.

Did you cover this with Nasir? When we render React JSX in any kind of loop, we have to specifically tell React what each element is... basically we have to give items IDs. You can check the [docs here](https://reactjs.org/docs/lists-and-keys.html#keys) for more info.

documentation/todo-app/src/components/todoItems.js
```diff
@@ -0,0 +1,22 @@
+import React, { Component } from 'react';
+
+export default class TodoItems extends Component {
+    render() {
+        const { description, deadline } = this.props;
+        
+        return (
+            <div> 
+                <tbody>
```
isaachinman: Why are you using `tbody` and `tr`, etc, here? First, if you want to use these tags, they need to be inside a `table` tag. Second, there's no reason to use them! HTML tables are pretty complicated/annoying things. You can definitely achieve everything you want with simple divs.

PR #100, opened 2018-05-03
Homework week3 hamza

zydan1:


isaachinman:
Seems like you've got layout issues at large resolutions:

screen shot 2018-05-04 at 13 10 41

PR #102, opened 2018-05-06
Emre3

arasmr:
I am sorry about my late commit. I was very busy in this week. I just finished it.


documentation/homework/week1/src/components/todoListItem.js
```diff
@@ -0,0 +1,24 @@
+import React,{Component} from 'react';
+
+class TodoListItem extends Component {
+    render() {
+        const id = this.props.id;
+        const description = this.props.description;
+        const deadline = this.props.deadline;
```
NouranMahmoud: You can use destructing here `const { id, description, deadline } = this.props;`

documentation/homework/week1/src/components/todoListItem.js
```diff
@@ -0,0 +1,24 @@
...
+        const done = this.props.done ? '◉' : '◎';
+        return (
+            <div className="mainContent">
+                <div className="content">
+                    {description}
+                     --->
+                    {deadline}
+                    <span className='doneClick' onClick = {()=> this.props.handleDoneClick(id)}>{done}</span>
```
NouranMahmoud: For writing good function names, you need to use real and meaningful verbs for what your function is doing. `handle` is bad practice to name your function but instead use **checkTodo** or **toggleTodo**.

documentation/homework/week1/src/components/todoList.js
```diff
@@ -0,0 +1,28 @@
...
+    render() {
+        const todoList = this.props.listTodo;
+        const todoListItem = todoList.map((data,index) =>
+             <TodoListItem 
+                id={data.id} 
+                description={data.description} 
+                deadline={data.deadline} 
+                done={data.done} 
+                handleDoneClick={this.props.handleDoneClick}
+                removeTodo = {this.props.removeTodo}
```
NouranMahmoud: It's not a good practice to put space between prop name or prop value and equal like `removeTodo = {this.props.removeTodo}` instead `removeTodo={this.props.removeTodo}`

documentation/homework/week1/src/components/todoList.js
```diff
@@ -0,0 +1,28 @@
...
+            />
+        );
+        return (
+            <div className='content'>
+                {todoListItem}
+                <TodoAdd addFunction= {this.props.addFunction}/>
```
NouranMahmoud: Don't you think that `addTodo` is far better? instead of `addFunction`

documentation/homework/week1/src/components/todoAdd.js
```diff
@@ -0,0 +1,40 @@
...
+
+class TodoAdd extends Component{
+    state = {
+        id:'',
+        description:'',
+        deadline:moment(),
+        done:false
+    }
+    onChangeState(field,value){
+        const currentState=this.state;
```
NouranMahmoud: Taking state in a variable like this, means you takes the reference of your state on `currentState` and changing anything using this reference will end up changing your original `state` itself (and we should NOT do this at all).

Instead you can rewrite this `onChangeState` like the follows

`this.setState((state) =>({...state, [field]: value}));`
or
`this.setState({...this.state, [field]: value})`

This expression `{...this.state}` takes all the properties on your state object and write it again in a totally new object.

If you don't know this syntax for es6 `{...this.state}` which ((destructuring assignment)) you can use `Object.assign({}, this.state, {[field]: value})`

**Note**: you also can use `this.setState({[field]: value})` without returning the full state.

documentation/homework/week1/src/App.js
```diff
@@ -0,0 +1,56 @@
...
+
+class App extends Component {
+  state={
+    listTodo:[],
+  }
+  addFunction(id,description,deadline,done) {
+    const currentState = this.state.listTodo;
+    const newState={id:newID(),description:description,deadline:deadline.format('YYYY-MM-DD'),done:done}
+    const updatedState=[...currentState,newState];
+    this.setState({listTodo:updatedState});
```
NouranMahmoud:
```
this.setState({
listTodo: [
  ...this.state.listTodo,
  {
    id: newID(),
    description,
    deadline: deadline.format('YYYY-MM-DD'),
    done
  }
]
});
```

Don't you think this is neater?, It's just a rewrite for your code but it's up to you how to write it :)

PR #105, opened 2018-05-06
Restructure homework submission

isaachinman:
@mkruijt Here's a PR with a rough approach as to what I would like for homework submission. All HW should go inside the `homework-submission` dir. Lint can be run via `yarn lint`, and that can be run automatically via CI (eg CircleCI) super easily, every time a PR is opened.


mkruijt:
@isaachinman should we wait to merge until the current class is done with the module?

isaachinman:
We can merge now, it doesn't make a difference as we never actually intend to merge any student's PR. This process can start with class 14. It doesn't really "work" without CI and (hopefully) appveyor, though.

PR #113, opened 2018-05-26
Sam's homework first week React

hackoor83:
Hello @isaachinman ,

This is the first week homework, with a quite good chunk of delay :)


isaachinman:
What is `App.1.js`? You should clean up your work before ever submitting a PR, so that it's very obvious to a reviewer what is happening in the code.

It seems like in `App.js`, many things are commented out. Why did you move the state (TodoJsonList) down your component tree into `ItemsList.js`, when it used to be in `App.js`? For React component state, you should pretty much never be moving state _down_, only up.

If you have specific questions please post them here and I will answer accordingly.

isaachinman:
It was not a review - we aren't going to review your work if you tell us from the beginning that it's broken.

I don't see any file named `ItemInput.js`. Please specify.

hackoor83:
Ah sorry, its a typo. I meant (InputForm.js).
@isaachinman

isaachinman:
First of all, if the function you're passing in as a prop to `InputForm` is called `addTodo`, you should call the prop `addTodo`, not something else (you called it `addItem`, this will just lead to confusion).

> I am really not sure how to pass the arrow functions with its arguments to the main App.js where the state resides

What exactly do you mean?

You're using refs in `InputForm.js`. We've already discussed how this is not the "React" way. You can store state either in `App.js`, or `InputForm.js`, and modify that state with an onChange function given to both your inputs.

Then to create the actual new todo, just use the state you already have.

All of this is covered explicitly and repeatedly both in docs and course material (of which there exists many videos on YouTube).

Let me know how else I can assist you. It seems like you need to spend some time reading more about how to handle forms in React in general.

isaachinman:
Hi Sam. As we discussed in class a few times, in the React world there are typically a thousand ways to achieve any task.

That being said, the issue of refs is not one of those cases. There is a clear and universal opinion from the entire community: don't use refs unless you need to.

Please check out the official docs on [Refs](https://reactjs.org/docs/refs-and-the-dom.html), and on [Uncontrolled Components](https://reactjs.org/docs/uncontrolled-components.html).

Directly from ReactJS docs:

> To write an uncontrolled component, instead of writing an event handler for every state update, you can use a ref to get form values from the DOM.

> Since an uncontrolled component keeps the source of truth in the DOM, it is sometimes easier to integrate React and non-React code when using uncontrolled components. It can also be slightly less code if you want to be quick and dirty. Otherwise, you should usually use controlled components.

What I am saying: it is certainly possible to create React forms with refs, and not use controlled components (inputs). But this is not advisable, and it is not how we teach the React module. If you want to submit your HW this way it is fine, but in order to get advice/help, we ask that you use controlled components so we can assist you more easily.

PR #123, opened 2018-06-07
Sarbast's week1 HW

SarKurd:


SarKurd:
Thanks for the feedback, i would remember that for future homeworks

PR #137, opened 2018-06-16
Murat hwweek2

MURATKAYMAZ56:
it is done:)


homework-submission/week2/src/App.js
```diff
@@ -0,0 +1,43 @@
...
+class App extends Component {
+  state = {
+    todos
+  };
+
+  onChange = event => {
+    const id = event.target.id;
+    const todos = this.state.todos;
+
+    todos.forEach(todo => {
```
NouranMahmoud: Using `forEach` to update array of object is a bad idea. Instead you should use *map*. The reason behind this is because *forEach* will loop over the array and update the todo object itself, however *map* function creates a new array and store the todos with the updated one on this newly created array and return it after the loop end.

Example:-
```
var newTodos = todos.map( todo => {
  if(todo.id === id) {
    return {...todo, done: !todo.done}
  }

return todo;
});

this.setState({ newTodos });
```

homework-submission/week2/src/App.js
```diff
@@ -0,0 +1,43 @@
+import React, { Component } from "react";
+import logo from "./logo.svg";
+import todos from "./todos.json";
+import Checkbox from "./Checkbox";
```
NouranMahmoud: It's better to name your component **Todo**, isn't it? :)

homework-submission/week2/src/Checkbox.js
```diff
@@ -0,0 +1,32 @@
...
+
+        return (
+            <li>
+                <input
+                    id={todo.id}
+                    type="checkbox"
+                    checked={todo.done}
+                    onChange={this.props.onChange}
+                />
+                <span style={todo.done ? doneStyle : unDoneStyle}>
```
NouranMahmoud: No need to create **unDoneStyle** you can use `null` instead or `{}`

PR #140, opened 2018-06-16
Sobhi HM Week2

Sob7i:


homework-submission/sobhiweek2/src/App.js
```diff
@@ -0,0 +1,51 @@
...
+    const todosState = this.state.todos;
+
+    const todosRender = todosState.map((elem) =>
+
+      <div key={elem.id}>
+       <span style={elem.done ? { textDecoration: 'line-through' } : { }}> 
+        <Description description={elem.description} />
+        <DeadLine deadline={elem.deadline} />
+        <input type="checkbox" className="checkbox"
+         onClick={this.handelCheck.bind(null, elem)}
```
NouranMahmoud: You are binding the `handleClick` with *elem* however you are not using any arguments in the function above. (line 13).  Also, you don't need to use *bind* in this case, instead call the function like this `onClick={() => this.handleCheck(elem)}. Because 'bind' may cause "performance issue" because this function create copies from this function. Read about 'bind' under the hood. :)

homework-submission/sobhiweek2/src/App.js
```diff
@@ -0,0 +1,51 @@
...
+      <div className="App" >
+
+        <Header /> 
+        <img className ="image" 
+              src="https://cdn-images-1.medium.com/max/2000/1*o-9xgkkyBgkhnBVN_Wxs6Q.png"/>
+         
+        <div className='todo-list'>
+
+          {todosRender} 
+          {(todosState.length === 0)? <h1>NO ITEMS</h1> : null}
```
NouranMahmoud: You can re-write this condition to be like this

```
{(todosState.length === 0)? <h1>NO ITEMS</h1> : todosRender} 
```

And remove the `{todosRender}` that's above it.

homework-submission/sobhiweek2/src/App.js
```diff
@@ -0,0 +1,51 @@
...
+import Header from "./source/Header";
+import Description from "./source/Description";
+import DeadLine from "./source/DeadLine";
+    
+
+export default class App extends React.Component {
+
+  state = { todos };
+
+  handelCheck = () => { this.setState({ todos: this.state.todos })};
```
NouranMahmoud: This should get a parameter and use it to update the todos list using the **map** function, Because it looks useless like this.

PR #142, opened 2018-06-16
Jason 2 week hw

thwjason:
Not sure if it works, because npm start results in crashes.


thwjason:
I have actually copied week 3's homework into this - I suppose if week 3 is an iterative process on week 2, that should suffice. 

Weird thing now is that it passes Travis here, but not in week 3. 

I am at a loss.

PR #160, opened 2018-06-25
homework week 3

hadeel43:


homework-submission/week3/src/App.js
```diff
@@ -0,0 +1,74 @@
...
+    const NewTodo = {
+      "id": this.state.data.length + 1,
+      "description": event.target.description.value,
+      "deadline": event.target.deadline.value,
+      "done": false
+    }
+    this.setState({ data: this.state.data.concat(NewTodo) });
+  }
+
+  removeItems = (i) => {
```
sarea: Not sure if this function works correctly!
I think this will always keep one element, however, I will not advice you to use `array.splice` in generally unless you really have to use it (which is very rarely)
but what you could do is filter out the element by it's id, that will return new array with new number of element and then you can check the length, if it > 0 show a message!
One last thing, I will highly recommend you to use good names for your variables/params 
for instance `i` is unreadable you could name it id, and so on!

PR #166, opened 2018-07-07
Fadi homework week5

Fadi-Rasheed:


JawharBairakdar:
Hi @Fadi-Rasheed.
### You made it, your homework `todoapp` main functionalities are working fine.
But I think you can make it better depending on best practice and for future better Code 🙂.

#### For the first look:
 - Fix your indentations (__important__)
 - you have to be more aware of using the decorators [Look Here](https://mobx.js.org/refguide/action.html#action)
>It is advised to use (@)action on any function that modifies observables or has side effects. action also provides useful debugging information in combination with the devtools.
 - you have to know your tools as well as you can, means you are fetching items with `PATCH` or `GET` or `DELETE` methods all of these are retuning a response and this response is a very useful thing we can deploy.

JawharBairakdar:
Hi @mkruijt Could you just keep this until I just review the updates as I promised @Fadi-Rasheed ? 🙂

homework-submission/todo-app/src/stores/TodoListStore.js
```diff
@@ -0,0 +1,107 @@
...
+        body: JSON.stringify(editTodo),
+        headers:{'content-type':'application/json', }
+      })
+        .then(res=> res.json())
+        .then(elem =>
+          runInAction(()=>{
+            item.description = elem.description;
+            item.deadline = elem.deadline;})) 
+        .then (()=> runInAction(()=>{
+          this.editing = null}))
```
JawharBairakdar: you can use `runInAction` for a once in the first `.then` method, so it would be like reseting and saving in the same time, in your case here you are sending a 2 different notifys to the store it self which will take more memory to update each one, which may take a long time on the UX *(user expreince)* in a large applications which means you will lose some of your users because of state updating 🙃, assuming you are working in amazon or a large company like it.

PR #168, opened 2018-07-08
add vanilla examples with mobx and react

NouranMahmoud:
These are 3 examples of the problem that mobx is solving
1. Vanilla app without mobx (the problem)
2. Vanilla app with mobx to solve the problem
3. React with Mobx.


isaachinman:
Hi @NouranMahmoud, maybe you can explain what this is for those of us that don't know?

It seems both vanilla examples don't use React. I guess the point here to show students the value of MobX? If so I would actually expect just to see a React vs React/MobX example, no? Comparing component state to MobX stores. Maybe it's better to demonstrate the value of React in separate examples, rather than combining the two things into one?

Just some thoughts!

NouranMahmoud:
Hi @isaachinman, These examples are for showing the need for Mobx (without react). So, first two examples are (vanilla-example, vanilla-mobx-example) shows the problem which is that we don't have a way to listen to changes that happen to properties of objects, then vanilla-mobx-example shows the solution using mobx which uses observer pattern (without decorators). The last one which is `react-mobx-example` turn/map the vanilla-mobx-example to a normal `react-mobx` project.

So the point here is not about React, it is all about Mobx usage. But we can add another example to demonstrate the value of React vs. Vanilla JS.

PR #171, opened 2018-07-10
Aya week4

Aya-Alabrash:


homework-submission/week4/src/components/TODOList.js
```diff
@@ -0,0 +1,29 @@
+/* eslint-disable */
+import React, { Component } from 'react';
+
+class TODOList extends Component {
+  render() {
+    const TODOList = this.props.TODOListItems.map((element, index) =>
+      <div key={index}>
```
JawharBairakdar: index is a good idea to use here but if you can use something else instead it would be better in avoiding Bugs cases

homework-submission/week4/src/store/TodoListStore.js
```diff
@@ -0,0 +1,49 @@
...
+  removeTodoItem = (itemId) => {
+    const newTodoItems = this.TODOListItems.filter((element) => element.id !== itemId);
+    this.TODOListItems = newTodoItems;
+  }
+
+  @action
+  addTodoItem = (e) => {
+    e.preventDefault();
+    const newTodo = {
+      "id": this.TODOListItems.length + 1,
```
JawharBairakdar: the Idea of `id` is to set a unique property for the object!

see? 🙃

homework-submission/week4/src/components/TODOList.js
```diff
@@ -0,0 +1,29 @@
...
+  render() {
+    const TODOList = this.props.TODOListItems.map((element, index) =>
+      <div key={index}>
+        <div onClick={() => this.props.checkStatus(element.id)} class="todoContent" >
+
+          {element.done ? <span>💙</span> : <span>🖤</span>}
+
+          <span style={element.done ? { textDecoration: 'line-through' } : {}}><strong>{element.description}</strong>, {element.deadline} </span>
+
+          <input type="submit" onClick={() => this.props.removeTodoItem(element.id)} value="Remove" name="remove" className="RemoveInput" />
```
JawharBairakdar: what if you seperated this to multy lines?

PR #173, opened 2018-07-11
Hw week4

WaelAlnadeem:


myapp/src/stores/TodoListStore.js
```diff
@@ -0,0 +1,63 @@
...
+                    ...element,
+                    done: !element.done
+                }
+            }
+            return element;
+        });
+        this.listItems = newTodoItems;
+
+    }
+    //    handleChangeCheckBox = (index) => {
```
JawharBairakdar: if no need remove it.
For best practice don't let the other developers confused with your code comments, and let your comments only for a useful information.

myapp/src/stores/TodoListStore.js
```diff
@@ -0,0 +1,63 @@
...
+
+configure({ enforceActions: true })
+
+
+class TodoListStore {
+    @observable listItems = todoListItem;
+
+    @action
+    handleChangeCheckBox = (id) => {
+        const newTodoItems = this.listItems.map((element) => {
```
JawharBairakdar: Short cut: 😃
```javascript
this.listItems = this.listItems.map(element => 
        id === element.id ? {...element, !element.done} : element
    )
```

myapp/src/stores/TodoListStore.js
```diff
@@ -0,0 +1,63 @@
...
+    //     const item = [...this.listItems]
+    //     item[index].done = !item[index].done
+    //     this.listItems[index].done = !this.listItems[index].done;
+    // }
+
+    @action
+    onDelete = (index) => {
+        const todosList = [...this.listItems]
+        todosList.splice(index, 1)
+        this.listItems = todosList;
```
JawharBairakdar: Nice use of `.splice` but you can also do it with `.filter`.
Like so with a short cut:
```javascript
this.listItems = this.listItems.filter((element, i) => i !== index)
```

JawharBairakdar: If you didn't get what is happening here mention me...

myapp/src/stores/TodoListStore.js
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @action
+    onAdd = (event) => {
+        event.preventDefault();
+        const newTodo = {
+            "id": this.listItems.length + 1,
```
JawharBairakdar: the Idea of `id` is to set a unique property for the object!

see? 🙃

myapp/src/stores/TodoListStore.js
```diff
@@ -0,0 +1,63 @@
...
+            "description": event.target.description.value,
+            "deadline": event.target.deadline.value,
+            "done": false
+        }
+        this.save(this.description, this.deadline, this.done, this.listItems.length)
+        this.listItems.push(newTodo);
+    }
+    @action
+    save = (description, deadline, index) => {
```
JawharBairakdar: I liked the Idea of `save` action 😃
you can do it shorter.
Can you figure it your self ? 🙂

myapp/src/Checkbox.js
```diff
@@ -0,0 +1,78 @@
...
+
+                </div>
+
+
+            );
+        }
+        else {
+            return (<div>
+                <form onSubmit={() => {
+                    this.props.save(this.state.description, this.state.deadline,
```
JawharBairakdar: `this.props.save` is undefined __!!__

PR #174, opened 2018-07-11
Week4

Rushkurdi:


homework-submission/myapp/src/App.js
```diff
@@ -0,0 +1,36 @@
+import React, { Component } from "react";
+import "./App.css";
+import Forms from "./components/forms";
+import TodoItems from "./components/TodoItems";
+import { inject, observer } from "mobx-react";
+import Header from "./components/header";
+
+@inject("RashadStore")
+@observer
```
JawharBairakdar: Depending on the Application size you have and the children levels it's Okay to `inject` on the top level of your app and pass it as props.

if you are in a real world Or your prject and because of a multy developers may work on the same file.
ONLY in case you have a one or two levels deep you can pass it other wise you are using MobX wasting your desk memory, and your team time following the code 🙃 .

homework-submission/myapp/src/components/forms.js
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+
+const Forms = ({ submitHandler }) => {
```
JawharBairakdar: Excellent way of use `props` 💯

homework-submission/myapp/src/components/header.js
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+import logo from "../logo.svg";
+
+class Header extends React.Component {
+  render() {
+    return (
+      <div className={this.props.className}>
```
JawharBairakdar: you hard coded it in `App.js` you can put it here simply 😄 instead of passing it as `props`

homework-submission/myapp/src/stores/rashadStore.js
```diff
@@ -0,0 +1,38 @@
...
+  @observable todos = Todos;
+
+  @action
+  checker = item => {
+    item.done = !item.done;
+  };
+
+  @action
+  remover = item => {
+    const remove = this.todos.filter(elem => elem !== item);
```
JawharBairakdar: Nice `.filter` use

homework-submission/myapp/src/stores/rashadStore.js
```diff
@@ -0,0 +1,38 @@
...
+    this.todos = remove;
+  };
+
+  @action
+  submitHandler = e => {
+    e.preventDefault();
+    const newTodo = {
+      id: this.todos.length + 1,
```
JawharBairakdar: the Idea of `id` is to set a unique property for the object!

see? 🙃

JawharBairakdar:
# it's not working
### Nice Work
for the first look to your application i figured some issuse you may have.

__Please make it works to make your work done__

PR #180, opened 2018-07-14
Sobhi week4

Sob7i:


JawharBairakdar:
Hi @Sob7i thank you for your homework.
### Please clean your branchs
You have a lot of files I couldn't figure out which one is the homework for each week, the homeworks should be in  `homework-submission/...` directory you have some there and one other in the repository it self called `sobhi_mobx/...`, also you have a branches for each week and that's Great But you have the homeworks weeks names in `homework-submission/...`

PR #195, opened 2018-08-09
added homework week1

neveenatik:


homework-submission/week1/src/App.js
```diff
@@ -0,0 +1,18 @@
...
+class App extends Component {
+  render() {
+    return <div className="App">
+        <header className="App-header">
+          <img src="https://russellinvestments.com/-/media/images/au/tiles/todolist.png" className="App-logo" alt="logo" />
+          <h1 className="App-title">Welcome to Todo List app</h1>
+        </header>
+        <p className="App-intro">This app is displaying three Todo items</p>
+        <TodoList />
+      </div>;
```
JawharBairakdar: I saw you are using this styling for more than a once which is fine. But I Highly recommend you to surround it with parentheses while every one else doing that not to be like them! but to make your code in other greater components more legible.
so you can just do it like so:
```javascript
class App extends Component {
  render() {
    return (
    <div className="App">
        <header className="App-header">
          <img src="https://russellinvestments.com/-/media/images/au/tiles/todolist.png" className="App-logo" alt="logo" />
          <h1 className="App-title">Welcome to Todo List app</h1>
        </header>
        <p className="App-intro">This app is displaying three Todo items</p>
        <TodoList />
      </div>
    );
  }
}
```

homework-submission/week1/src/TodoList.js
```diff
@@ -0,0 +1,13 @@
+import React, { Component } from 'react';
+import todoItems from './todoItems.js';
+
+class TodoList extends Component {
+
+  todoItem = todoItems.map((item, i) => <li key={i}>{ "Task: " + item.description + " deadLine: " + item.deadLine}</li>);
```
JawharBairakdar: Here too I think here is a bit more important than the previous one.
so if like I think it would be better for readers more than a long nice line: :slightly_smiling_face: 
```javascript
todoItem = todoItems.map((item, i) => (
    <li key={i}>
      { "Task: " + item.description + " deadLine: " + item.deadLine}
    </li>
  ));
```

homework-submission/week1/src/App.js
neveenatik: Thanks a lot for your review and feedback I made the changes.

homework-submission/week1/src/TodoList.js
neveenatik: Indeed it is more readable. Changes are made.

PR #196, opened 2018-08-09
week1 HW - Meazer

Meazer:


sarea:
Very nice work!
I like that you added json file for the data! it is really nice to simulate the api call!
also I like your code structure! separate component for each thing! like it!

if you see you have the ability to learn more, I will advice you to learn about the component in deep,
- `Dumb Components and Smart Components` https://medium.com/@thejasonfile/dumb-components-and-smart-components-e7b33a698d43
- `Class Component vs function Component`  https://medium.com/@paulrohan/understanding-reactjs-component-741f3f0e4a8f

PR #198, opened 2018-08-10
Dumie Homework week1

Dumie1:


homework-submission/homework-week1/src/App.js
```diff
@@ -0,0 +1,22 @@
...
+    return (
+      <div className='App'>
+        <header className='App-header'>
+          <img src={logo} className='App-logo' alt='logo' />
+          <h1 className='App-title'>Todo List</h1>
+        </header>
+        <p className='App-intro'></p> 
+        <Description descriptionAndDeadline='Get out of bed , Wed Sep 13 2017' />
+        <Description descriptionAndDeadline='Brush teeth , Thu Sep 14 2017' />
+        <Description descriptionAndDeadline='Eat breakfast , Fri Sep 15 2017' />
```
JawharBairakdar: you almost there This is almost the right way using the Component, you have some main points here to mention:
 1. you used one property of `props` to pass into the component and you should separate your `date`s from your texts always.
 2. the naming is not about the description it self so it's a `Description` yes, but about what I meant what Kind of it you are doing here? what about `TodoItem`? it should be enough describing what is contains, But I still think you will have a better name!!
 3. last thing I saw that your HTML looks like so:
```javascript
<div>
  <ul>
    <li>Get out of bed , Wed Sep 13 2017 </li>
  </ul>
</div>
<div>
  <ul>
    <li>Brush teeth , Thu Sep 14 2017 </li>
  </ul>
</div>
<div>
  <ul>
    <li>Eat breakfast , Fri Sep 15 2017 </li>
  </ul>
</div>
```
according to this HTML structure you may not be able to assign a new Item as an `li` tag, so even if you have a very dynamic way of structuring the `Description` Component here this will only repeat it self more and more.

what do you have to do is just moving the `div` and the `ul` out side of the component so you will have more control over the component in your case :slightly_smiling_face:

Dumie1: Thank you for the constructive feedback. I will fix my code according to the feedback.

PR #204, opened 2018-08-22
Week2 mahmoud

AllouhM:


homework-submission/week2/src/component/Todolist.js
```diff
@@ -0,0 +1,23 @@
+import React, { Component } from "react";
+import Todo from "./Todo";
+
+class Todolist extends Component {
+  render() {
+    const { items } = this.props;
+    const todos = items.map(todo => (
```
JawharBairakdar: Actually I spent some extra time to understand hoe this is working.
So What is `todos` or `items` in other words? where it came from?
Answering this question `todos` is an object it came from a `json` has been passed through state then props file that contains an array of objects.

That's right ! But the issue is not with todos, The Issue is how react is working, and by setting an object out sid of the react component is not a solution of making things works, so it may work here and it's working very well I Agree with that, But in the near future you will see that so clearly it's not the way of working with react.

**React is Reacting on notifications that are Coming from state or props**

So how the state or the props are behave, it's working.

in your scenario why is't working right ?! 
well the answer is so simple, you are changing a one singular element and the element is reacting on what you are giving him, so not react it's JSX engine.

So Now you have to follow this in order to understand what is going on...
I need you before you read the point number two I really want you to go where the number one is telling you to go.
 1. State is your component store, and setState is the component notifications manager with props passed from the parent, unless you have something like MOBX or Redux, [GO Here](https://github.com/HackYourFuture/React/pull/204/files#diff-8d80e97daabceb39a2dead2cdd28e3ceR9).
 2. for using `data` what do you have to do is just to change the `handleChange` method 🙂.
 3. now we know from where the problem started and we should also know where it's [going!!]https://github.com/HackYourFuture/React/pull/203/files#diff-dd04c95323f0527b8ead2f8aa8237f4bR18)

homework-submission/week2/src/App.js
```diff
@@ -0,0 +1,34 @@
...
+class App extends Component {
+  state = {
+    data
+  };
+
+  handleChange = todo => {
+    const todos = this.state.data;
+    todo.done = !todo.done;
+    this.setState({ todos });
+  };
```
JawharBairakdar: By looking to this method line 14 exactly for a vanilla javascript it may has it own effect.

But in react as your application it has it own effect out side of react _**JSX effect**_ and what you are doing here simply is just resetting the `state` as it's without making any changes on it each time you are making a change on the checkbox.

so what is happening that you are using the `setState` outwardly for making a fake effect on the **view** which is the render methodology. _(simply the notification stuff we talked about them)_
what you are really changing is the object you grabbed from the `json` file on the flow.

so now after you make your previous changes there what you can do here is:
 - there is no need for line 14 because as we said it's just changing the content on the flow, and in case of state you are changing the state manually which we should be away of it.
 - you have to loop through the `this.state.data` using map.
 - inside of `.map` you should for the first step check for the `id` when it matches the `todo.id` with an object of the `todos` objects you should return the object with the new content, otherwise return the same object.

also I should mention that you are setting a new property on your state without any reason or use.
 15. we have a very duplicated array with a different name:

It's works fine but you got in result:
```javascript
// from line 15
// The expected State with some changes internally
Object { data: (3) […] }

// The new State we are getting
Object { data: (3) […], todos: (3) […] }
```
How we can solve this is just to think somehow simpler as much as we can !!
 - in line 15 you are setting the state like so:
```javascript
 // this line telling React about an object it existed to create a new item inside of it with the name: `todos`
this.setState({ todos });

// the correct way to use it is like so:
this.setState({ data: todos }); // it should work
```

As I told it's not the very correct way to insert the whole thing for modifying a one item even if you have a very small object.
but also it's a better practice to have a full control over the whole thing instead.
So what I could do here is using `.map` 
 1. it will make a new copy for me that will prevent any changes to the state.
 2. it will loop through them a one by one giving me a better control over each!!
 3. it's more readable.

for more about es6 the spreed operator I think I have a really nice and short article to [read](https://medium.com/@nothingisfunny/javascript-spread-syntax-8e8663938a1c).

homework-submission/week2/src/App.js
```diff
@@ -0,0 +1,34 @@
+import React, { Component } from "react";
+import Todolist from "./component/Todolist";
+import "./App.css";
+import "./todoapp.css";
+import data from "./component/data.json";
+
+class App extends Component {
+  state = {
+    data
```
JawharBairakdar: You are setting `data` as a state with no real actual use of it !!

homework-submission/week2/src/component/Todolist.js
```diff
@@ -0,0 +1,23 @@
...
+      <Todo
+        todo={todo}
+        key={todo.id}
+        handler={() => this.props.handler(todo)}
```
JawharBairakdar: why are you calling it? and why are you inserting it inside of another object ?

PR #212, opened 2018-08-27
hw w1

amudar:


homework-submission/react-hw/src/App.js
```diff
@@ -0,0 +1,25 @@
...
+import Description from "./components/Description";
+
+class App extends React.Component {
+  render() {
+    return (
+      <div>
+        <ul className="list">
+          <h1>ToDo Checklist</h1>
+          <li>
+            <Description todo={"Get out of bed"} deadline={"Wed Sep 13 2017"} />
```
JawharBairakdar: very nice passing to the props, you are in the future now so I won't tell that you could do it with an object instead 😅

amudar:
Hi @JawharBairakdar , Thank you for the feedback. Error is fixed, please check it now.

PR #213, opened 2018-08-28
hw w2

amudar:


homework-submission/react-hw/src/App.js
```diff
@@ -0,0 +1,57 @@
...
+  state = {
+    todoList
+  };
+
+  clickHandler = e => {
+    const tempArr = this.state.todoList.slice();
+
+    console.log(e.target.id);
+
+    tempArr[e.target.id - 1].done = !tempArr[e.target.id - 1].done;
```
JawharBairakdar: You have a very nice way of updating the state here, But you are modifying the state immediately, you can test this instead of yours and you will find out that you are doing it:
```javascript
const tempArr = this.state.todoList.slice();

console.log(this.state.todoList[e.target.id - 1].done); // this will show you the previous state
tempArr[e.target.id - 1].done = !tempArr[e.target.id - 1].done;
console.log(this.state.todoList[e.target.id - 1].done); // this will show you the next state before setting it !!
```
So it would be better to use for a `.map` method here that doing all of these for you and it's just more readable for others.
So you may make it as
```javascript
this.setState({
  list: this.state.todoList.map(item => {
    if (item[id] === id ){ // you have to get the id from somewhere !!  -- you can play around it 😁
      // return a new Object that contains the new data we need
      return {
        ...item, // using the spread operator we will copy the elements inside of `item` and replace the `done` with a new value
       done: !item.done
      }
    }
    return item
  })
});
```
Or you may just do the following:
```javascript
// replacing the line 16 with this
tempArr[e.target.id - 1] = {
   ...tempArr[e.target.id - 1], done: !tempArr[e.target.id - 1].done
 }
```
and if you get what is happening here you should remove another line too.

PR #217, opened 2018-09-02
Husam - homework week 3

ThreePalmTrees:
updated homework after feedback,
please request changes if I have to change something


JawharBairakdar:
Hi @ThreePalmTrees, I think there is something wrong with this `@vaadin/vaadin-date-picker` madule.
I will wait for a while and re-reviewed again if there is any technical issues they have.

ThreePalmTrees:
Thanks @JawharBairakdar 🙌

homework-submission/src/Components/UserAction.js
```diff
@@ -0,0 +1,39 @@
...
+function Cancel(props) {
+  const { handleCancel } = props;
+  return (
+    <button className="btn" onClick={handleCancel}>
+      Cancel
+    </button>
+  );
+}
+
+export { Edit, Update, Remove, Cancel };
```
keepitterron: I like it! 💯

homework-submission/src/Components/list.css
```diff
@@ -0,0 +1,104 @@
...
+    rgba(15, 15, 15, 0.151)
+  );
+  margin-top: 25px;
+}
+
+.list li {
+  padding: 10px;
+  padding-top: 32px;
+  padding-left: 25px;
+  padding: 15px;
```
keepitterron: we're having a little padding party here 🎉 
what do you think will happen here? Which padding will be applied?

homework-submission/src/Components/list.css
```diff
@@ -0,0 +1,104 @@
...
+
+.list {
+  display: flex;
+  flex-direction: column;
+  list-style-type: none;
+  width: 600px;
+  max-height: 70vh;
+  overflow-y: scroll;
+  margin: 50px auto;
+  padding: 50px 80px 70px 60px;
```
keepitterron: ⚠️ this is an antipattern. The padding should always be consistent and harmonic. You have different values for each side, this means the design is wrong or the implementation is wrong. 90% of the case the padding should be the same for top/bottom and left/right, otherwise your component will look weird.

Why are you using so many different sizes?

homework-submission/src/Components/updateTodo.css
```diff
@@ -0,0 +1,64 @@
+#enter-new-description {
```
keepitterron: Is there a reason why are using ids here and not classes?
I always try to use classes and very rarely use ids, [you can find out why in this article](https://dev.to/claireparker/reasons-not-to-use-ids-in-css-4ni4)

homework-submission/src/App.js
```diff
@@ -0,0 +1,206 @@
+import React, { Component } from "react";
+import "./App.css";
+import todosJSON from "./Sources/todosJSON.json";
+import List from "./Components/List";
+import NewTodoForm from "./Components/NewTodoForm";
+import uuid from "uuid/v4";
```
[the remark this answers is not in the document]
ThreePalmTrees: Weird, I installed it using `npm i uuid` 
Maybe I added the wrong package.json file, I'll fix that right away 👍

homework-submission/src/App.js
```diff
@@ -0,0 +1,206 @@
...
+      description: "",
+      deadline: "",
+      id: ""
+    },
+    actions: {
+      editClicked: false,
+      updatedTodo: {
+        description: "",
+        deadline: "",
+        itemID: ""
```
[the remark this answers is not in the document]
ThreePalmTrees: Maybe it's not the best way to name things, but it made sense to me when I was doing it, 
The `newTodo` is supposed to be the new submitted todo, and the `updatedTodo` is for storing when the user wants to edit an existing todo, the `id` & `itemID` are the same, but everyone handles a different situation, is this a good way to do it ? I guess I can change the name of `id` to be consistent with what I'm naming the other id's

homework-submission/src/App.js
```diff
@@ -0,0 +1,206 @@
...
+      }
+    }
+  };
```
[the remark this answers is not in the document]
ThreePalmTrees: That makes sense lol, 
Gonna do as you suggested :)

homework-submission/src/App.js
```diff
@@ -0,0 +1,206 @@
...
+      if (todo.id === item.id) {
+        return { ...todo, done: !todo.done };
+      }
+      return todo;
+    });
+
+    // ✓
+    // console.log(this.state.todos, todosCopy);
+    this.setState({ todos: todosCopy });
+  };
```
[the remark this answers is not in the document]
ThreePalmTrees: Was on autopilot :D
Yes that makes perfect sense,
Gonna make it return only the todo with the corresponding id and gonna name the function `toggleDone` 👍

homework-submission/src/App.js
```diff
@@ -0,0 +1,206 @@
...
+
+  handleNewDescription = e => {
+    let { newTodo } = this.state;
+    newTodo = { ...newTodo, description: e.target.value };
```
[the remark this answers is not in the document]
ThreePalmTrees: This must have slipped when I was refactoring, gonna fix it right away

homework-submission/src/App.js
```diff
@@ -0,0 +1,206 @@
...
+      description: "",
+      deadline: "",
+      id: "",
+      done: false
+    };
+
+    // ✓
+    // console.log(this.state.todos, todosCopy);
+    this.setState({ todos: todosCopy, newTodo: newTodo });
+  };
```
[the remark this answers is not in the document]
ThreePalmTrees: Yes that looks cleaner and more readable, will implement that

homework-submission/src/App.js
```diff
@@ -0,0 +1,206 @@
...
+    // console.log(this.state.todos, newTodos);
+    this.setState({ todos: newTodos });
+  };
+
+  handleEdit = itemID => {
+    const { actions, todos } = this.state;
+    let actionsCopy = { ...actions, editClicked: true };
+    actionsCopy.updatedTodo = { ...actionsCopy.updatedTodo, itemID };
+
+    const todo = todos.filter(todo => todo.id === itemID)[0];
```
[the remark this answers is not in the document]
ThreePalmTrees: I did this in the week4

homework-submission/src/App.js
```diff
@@ -0,0 +1,206 @@
...
+    actionsCopy.updatedTodo = {
+      ...actionsCopy.updatedTodo,
+      description,
+      deadline
+    };
+
+    // ✓
+    // console.log(this.state.actions.updatedTodo, actionsCopy.updatedTodo);
+    this.setState({ actions: actionsCopy });
+  };
```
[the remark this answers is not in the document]
ThreePalmTrees: Way shorter and better in my opinion, Gonna change the code too 👍

homework-submission/src/App.js
```diff
@@ -0,0 +1,206 @@
...
+          dataModel={this.state.todos}
+          handleCheck={this.handleCheck}
+          handleEdit={this.handleEdit}
+          handleRemove={this.handleRemove}
+          handleUpdate={this.handleUpdate}
+          actions={this.state.actions}
+          handleUpdateDescription={this.handleUpdateDescription}
+          handleUpdateDeadline={this.handleUpdateDeadline}
+          handleCancel={this.handleCancel}
+        />
```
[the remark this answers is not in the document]
ThreePalmTrees: To be honest I was saying that to myself but I couldn't think of a better way to to that without a state management tool

homework-submission/src/App.js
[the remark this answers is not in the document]
keepitterron: Keep in mind that all the code examples i'm giving you are not tested 😝 
It's important to understand how the examples I give you make the code better and - most importantly - what the code does. sometimes it's just conventions, they don't make the code better but just easier to understand for other developers.

homework-submission/src/App.js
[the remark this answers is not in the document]
keepitterron: this code is fine, it does what the assignment asks, and you should be proud of yourself for realising it can be improved!
If you want to watch a fun video on how to make things simpler in React, I highly suggest [Simply React from Kent C. Dodds](https://www.youtube.com/watch?v=AiJ8tRRH0f8)

homework-submission/src/Components/list.css
ThreePalmTrees: Maybe I'm treating css as if it was Photoshop, 
I think I'll learn bootstrap after we finish the final project, or I'll learn proper design principles

homework-submission/src/Components/list.css
ThreePalmTrees: The last one 15px will be applied, sorry I didn't refactor the css code well

homework-submission/src/Components/updateTodo.css
ThreePalmTrees: Yes I try to use classes too whenever I can, unless I know this element is special and its style won't be replicated
Was operating on autopilot I guess 😀, thank you so much Claudio, I'm working on my homework and following your feedback, which was very helpful 👐

homework-submission/src/App.js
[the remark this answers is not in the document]
ThreePalmTrees: You're a saint 
Thank you for the video 👐

homework-submission/src/Components/updateTodo.css
ThreePalmTrees: @keepitterron I kept using id instead of class because the styles weren't being applied for some reason

homework-submission/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "reactweek3",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@vaadin/vaadin-date-picker": "^3.2.0",
+    "react": "^16.4.2",
+    "react-dom": "^16.4.2",
+    "react-scripts": "1.1.4",
+    "uuid": "^3.3.2"
```
keepitterron: 🎉

homework-submission/src/App.js
```diff
@@ -0,0 +1,136 @@
+import React, { Component } from "react";
+import "./App.css";
+import todosJSON from "./Sources/todosJSON.json";
+import List from "./Components/List";
+import NewTodoForm from "./Components/NewTodoForm";
+import uuid from "uuid/v4";
+
+const TODO_TEMPLATE = {
```
keepitterron: 🎉

homework-submission/src/App.js
```diff
@@ -0,0 +1,136 @@
...
+
+  handleNewDescription = e => {
+    const newTodo = { ...this.state.newTodo };
+    newTodo.description = e.target.value;
+
+    this.setState({ newTodo });
+  };
+
+  handleNewDeadline = e => {
+    // is using a ref better here ? then combining this and handeleNewDescription in one go
```
keepitterron: there are multiple techniques you can use to make this code re-usable but it would require an architectural change. I think it looks already very well :)

homework-submission/src/App.js
```diff
@@ -0,0 +1,136 @@
...
+  };
+
+  handleSubmit = e => {
+    e.preventDefault();
+    let newTodo = { ...this.state.newTodo };
+    if (!newTodo.description || !newTodo.deadline) return;
+
+    let { todos } = this.state;
+    newTodo = { ...newTodo, id: newTodo.id || uuid() };
+    todos = [...todos, newTodo];
```
keepitterron: you are saving a reference to `state.todos` and then mutating it here, I'm sure you just got caught up in refactoring :)
you can simply do `const todos = [...this.state.todos, newTodo]` without the intermediate [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)

homework-submission/src/App.js
```diff
@@ -0,0 +1,136 @@
...
+
+  handleRemove = itemID => {
+    const { todos } = this.state;
+    const newTodos = todos.filter(todo => todo.id !== itemID);
+
+    this.setState({ todos: newTodos });
+  };
+
+  handleEdit = itemID => {
+    let updatedTodo = { ...this.state.updatedTodo };
```
keepitterron: you are copying `updateTodo` from state - good! - but then you are overwriting it 3 lines below with `updateTodo = entity`. Is there a reason I don't see?
Instead of calling it `entity` you can call it directly `updateTodo` there, `find` returns a copy of the object, it's immutable by default 🎉 

It would become:
```js
const todos = [...this.state.todos];
const updatedTodo = todos.find(todo => todo.id === itemID);
this.setState({ updatedTodo });
```

homework-submission/src/App.js
ThreePalmTrees: indeed, got caught up in refactoring :)
fixed it though, thanks Caludio

homework-submission/src/App.js
ThreePalmTrees: i don't know why i over complicate things, must do a zen meditation before starting to code 😄

keepitterron:
ahahah, no torture at all :) I'm happy to help!

What I would do in this case to clean up the dependencies and reduce the props passed is indeed let some components handle their own state, it would keep track of description and deadline internally and call a function to update the parent's state with a `todo` object when the user clicks save.
The only source of truth would be the `App` component `state.todos`, but I see no reason for this high level component to also know about the editing state, that can be handled by another component :)

PR #221, opened 2018-09-04
Homework week3

zmagzoub:
Sorry because I am late. Bonus homework 'edit todo' is not done. I will do it in week4 homework.


homework-submission/week1/src/App.js
```diff
@@ -0,0 +1,65 @@
...
+import AddTodo from './components/AddTodo';
+import data from './sources/data.json'
+
+class App extends Component {
+  state = {
+    data
+  }
+    handler = e => {
+      let todos = this.state.data;
+      todos[e.target.id -1].done = !todos[e.target.id -1].done;
```
NouranMahmoud: This way you are mutating the todos **directly** which is the behaviour that React is trying to avoid!.

Taking the `todos` in a variables like this `let todos = this.state.data;` doesn't create a new copy from `data` but instead it creates a new (pointer) to the same place where we store `data`. Based on this, line 14 will be updating the state **directly**.

What we need to do is to take a copy from our state which is **data** and update this copy without touching the original copy that we have in our state. Then, we ask the `this.setState` to update the state for us, which is the intended behaviour from React!.

There're many ways to take a copy from an object or array, one of them is `let todos = [...this.state.data]` (es6 syntax) and another way is `let todos = this.state.data.map( (item) => item);` and `let todos = this.state.data.slice()` and many other ways that you search about :) ..

However, In this case it is better to . use **filter**

```
this.state.data.map( (todo, i) => (
   {...todo, done: (e.target.id -1 === i) ? !todo.done : todo.done }
))
```

Please if this not clear, don't hesitate to contact me on slack to clarify it more :)

homework-submission/week1/src/App.js
```diff
@@ -0,0 +1,65 @@
...
+      data.push(newTodo)
+      this.setState({
+        data
+      })
+      e.target.description.value = "";
+      e.target.deadline.value = "";
+    }
+
+    removeTodo = (i) => {
+      this.state.data.splice(i, 1);
```
NouranMahmoud: The same here, you are deleting the todo from the original state!. Please fix this.

homework-submission/week1/src/App.js
```diff
@@ -0,0 +1,65 @@
...
+    addTodo = e => {
+      e.preventDefault();
+      const {data} = this.state;
+      const newTodo = {
+        id: data.length + 1,
+        description: e.target.description.value,
+        deadline: e.target.deadline.value,
+        done: false
+      };
+      data.push(newTodo)
```
NouranMahmoud: The same here, you are pushing directly to the original state!, please fix this.

homework-submission/week1/src/App.js
```diff
@@ -0,0 +1,65 @@
...
+      this.setState({
+        todos
+      })
+    }
+
+    addTodo = e => {
+      e.preventDefault();
+      const {data} = this.state;
+      const newTodo = {
+        id: data.length + 1,
```
NouranMahmoud: It's better to generate random id whether by using a module to do this (there're many options, one of them are `uuid`) or to generate random number by `Math.random()`

homework-submission/week1/src/components/RemoveTodo.js
```diff
@@ -0,0 +1,11 @@
+import React, { Component } from 'react';
+
+class RemoveTodo extends Component {
+    render(){
+        return(            
+                <button onClick= {()=>{this.props.remove(this.props.i)}} type='submit'>Remove</button>
```
NouranMahmoud: No need for `type="submit"` here, because this button is not a part of a `<form>`.

homework-submission/week1/src/components/Todos.js
```diff
@@ -0,0 +1,34 @@
...
+              checkbox = {item.done} 
+              handler = {this.props.handler}
+             />
+
+            <List 
+              description = {item.description} 
+              deadline = {item.deadline}
+             />
+             <RemoveTodo
+             remove = {this.props.remove}
```
NouranMahmoud: There's a better way to bind your `this.props.remove` function with the **id** without passing it to the `<RemoveTodo>`

```
<RemoveTodo
  remove={this.props.remove.bind(null, i)}
>
```

This `this.props.remove.bind(null, i)` will return a new function with `id` already stored inside it.

The `.bind` take the `id` store this id (somehow) until the function get called then it send the id by its own.

You can read about `bind` and what happens under the hood.

zmagzoub:
Thank you for your review Nouran. I've updated my homework.

PR #222, opened 2018-09-05
Knar -Homework week 4

PKnar:
Like homework instruction suggested. I decided to keep 'TodoItem' and 'Deadline' components' state as a local state and use MobX for parent (App) component state only.


homework-submission/week4-homework/src/App.js
```diff
@@ -0,0 +1,49 @@
...
+        onRemove={() => {
+          TodoStore.deleteTodo(todo.id);
+        }}
+        onUpdate={TodoStore.updateTodo}
+      >
+        <Checkbox
+          key={index}
+          done={todo.done}
+          onClick={() => {
+            TodoStore.toggleDone(todo.id);
```
JawharBairakdar: this line is a very nice implementing of a function call with passing in a short way instead of moving it to the side of the child component, But it's better to note about that in the comment in the child component.

homework-submission/week4-homework/src/components/Checkbox.js
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+class Checkbox extends React.Component {
+  render() {
+    let done = this.props.done;
+    return (
+      <button onClick={this.props.onClick} className="checkbox">
```
JawharBairakdar: Usually when we see this line of code we say that it's mentioned to an event Object passed by an argument field, That I didn't mention before but it's good to know.

So if you want to pass something pass it what may make more sense for you in the future and for others as well.

PKnar:
Hi, @JawharBairakdar  thank you very much for your feedback! :+1:  in my previous homework my function was called handleClick but because it was not clear for you what that function did when you were reviewing it, I thought maybe naming it toggleDone would make it more clear.

PR #225, opened 2018-09-08
week4-hw

amudar:


homework-submission/react-hw4/src/components/stores.js
```diff
@@ -0,0 +1,39 @@
+import { observable, action, computed, configure } from "mobx";
+import Todos from "./todos.json";
+
+configure({ enforceActions: true });
+
+class Stores {
```
NouranMahmoud: Why **Stores**, it is only one store?

homework-submission/react-hw4/src/components/stores.js
```diff
@@ -0,0 +1,39 @@
+import { observable, action, computed, configure } from "mobx";
```
NouranMahmoud: It's better to put the store in a folder called **stores/**

homework-submission/react-hw4/src/components/stores.js
```diff
@@ -0,0 +1,39 @@
...
+import Todos from "./todos.json";
+
+configure({ enforceActions: true });
+
+class Stores {
+  @observable
+  todos = Todos;
+
+  @action
+  checker = item => {
```
NouranMahmoud: This action doesn't update any observable in the store!?, so why did you create it?

homework-submission/react-hw4/src/components/stores.js
```diff
@@ -0,0 +1,39 @@
...
+    item.done = !item.done;
+  };
+
+  @action
+  remover = item => {
+    const remove = this.todos.filter(elem => elem !== item);
```
NouranMahmoud: Give meaningful names to your variables/functions/classes!, `const remove` can be `const todosList ...`

NouranMahmoud: You should not compare `elem, item` because they are full objects .. so in this case you are comparing something like this `{} === {}` which is **always false**.. you should compare using primitive values within these objects, like `elem.id === item.id`.

homework-submission/react-hw4/src/components/stores.js
```diff
@@ -0,0 +1,39 @@
...
+    this.todos = remove;
+  };
+
+  @action
+  submitHandler = e => {
+    e.preventDefault();
```
NouranMahmoud: Keep any event handling on the component, but in store just use naming the reflect the action that should be send to the backend, like `createTodo` instead of `submitHandler`

homework-submission/react-hw4/src/components/stores.js
```diff
@@ -0,0 +1,39 @@
...
+    const newTodo = {
+      id: this.todos.length + 1,
```
NouranMahmoud: This way of generating ids is error prone, so please use instead the `uuid` npm module or `Math.random()`

homework-submission/react-hw4/src/App.js
```diff
@@ -0,0 +1,30 @@
...
+          todos={todos}
+          remover={remover}
+        />
+        {count}
+      </div>
+    );
+  }
+}
+
+export default observer(App);
```
NouranMahmoud: You don't need observer(App), because you already using the decorator @observer

PR #227, opened 2018-09-11
Homework week4

zmagzoub:
Sorry because I am late.


homework-submission/week1/src/components/Checkbox.js
```diff
@@ -0,0 +1,16 @@
+import React, { Component } from 'react';
+
+class Checkbox extends Component {
+    render(){
+        return(
+            <input 
+                type = 'Checkbox' 
+                defaultChecked = {this.props.item.done} 
+                id = {this.props.id}
```
JawharBairakdar: Perhaps you could have another look at the `id` property?

homework-submission/week1/src/components/List.js
```diff
@@ -0,0 +1,14 @@
+import React, { Component } from "react";
+
+
+class List extends Component {
+    render() {
+        return (
+            <div>
+                <li>{this.props.description}, {this.props.deadline} </li>
```
JawharBairakdar: you are inserting an `li`inside of a `div` tag, from the [documentations](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/li) we will notice that the Permitted parents fro an element are very specific, it's okay for an example but in a real world it may cause a bugs.

homework-submission/week1/src/components/Todos.js
```diff
@@ -0,0 +1,33 @@
...
+import RemoveTodo from "./RemoveTodo";
+
+class Todos extends Component {
+  render() {
+    return (
+      <div>
+        <ul>
+          {this.props.items.map((item, id) => (
+            <div key={id} className={item.done? 'checked' : ''}>
+            <Checkbox
```
JawharBairakdar: Here too, you have a `div` inside of a `ul` which will make the following tree with what is inside the List Component:
```HTML
<ul>
  <div>
    <div>
      <li>
      </li>
    </div>
  </div>
</ul>
```

homework-submission/week1/src/components/Todos.js
```diff
@@ -0,0 +1,33 @@
...
+              item = {item}
+             />
+
+            <List 
+              description = {item.description} 
+              deadline = {item.deadline}
+             />
+             <RemoveTodo
+             remove = {this.props.remove}
+             i = {id}
```
JawharBairakdar: Here I would call this a more descriptive name such as `todoIndex` because we are using the index.

PR #228, opened 2018-09-12
Week3

amirno1:
Due to some situation i had no time to hand in homework week 3 and 4. I am trying my best to catch up. thanks for your feedback in advance.


homework-submission/reactweek1/src/App.js
```diff
@@ -0,0 +1,123 @@
...
+        },
+      ],
+      newDescription: '',
+      newDeadline: ''
+    }
+  }
+
+  changeToDone = (event) => {
+    const id = event.target.id;
+    const listCopy = this.state.list.slice(0);
```
keepitterron: nice! you can also use `[...this.state.list]`, it's the same thing but shorter :)

homework-submission/reactweek1/src/App.js
```diff
@@ -0,0 +1,123 @@
...
+      list: listCopy
+    });
+  }
+  descriptionHandler = (newDescription) => {
+    const state = this.state;
+    state['newDescription'] = newDescription;
+    this.setState(state)
+  };
+
+  deadLineHandler = (newDeadLine) => {
```
keepitterron: same as above 🐕

homework-submission/reactweek1/src/App.js
```diff
@@ -0,0 +1,123 @@
...
+    const state = this.state;
+    const newItem = {
+      'id': this.state.list.length + 1,
+      'description': this.state.newDescription,
+      'deadline': this.state.newDeadline,
+      'done': false
+    };
+    this.state.list.push(newItem);
+    // this.state.newDeadline = '';
+    // this.state.newDeadline = ''; commented out these two cus it faces some problem through the test but dont know what the better way is
```
keepitterron: it's a minor mistake and it's very common :)
you are using `this.state.newDeadline` twice, i think the second one is supposed to be `this.state. newDescription `?

Here you are mutating the state again :)
You can avoid that by using [the spread operator](https://zendev.com/2018/05/09/understanding-spread-operator-in-javascript.html):
```js
this.setState({list: [...state.list, newItem]});
```
What do you think?

homework-submission/reactweek1/src/App.js
```diff
@@ -0,0 +1,123 @@
...
+    this.setState(state)
+  };
+
+  removeHandler = (i) => {
```
keepitterron: can you also change this function not to mutate the state?
hint: you can use [Array.prototype.filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter)

homework-submission/reactweek1/src/App.js
```diff
@@ -0,0 +1,123 @@
...
+          })
+        }
+      </ul>
+      <label>
+        Enter description :
+              <input type="text" onChange={(e) => this.descriptionHandler(e.target.value)} value={this.state.list.newDescription} />
+      </label>
+      <label>
+        Enter deadline :
+              <input type="text" onChange={(e) => this.deadLineHandler(e.target.value)} value={this.state.list.newDeadline} />
```
keepitterron: There is a major problem here. The user can insert whatever they want here :) How can we make sure that they only insert a date formatted how we want it?
hint: [take a look at the `date` input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date)

homework-submission/reactweek1/src/App.js
amirno1: I think that you are absolutely right, but just for my information what does '...' mean in this line before state.list?

homework-submission/reactweek1/src/App.js
```diff
@@ -0,0 +1,123 @@
...
+    const itemIndex = listCopy.findIndex(el => { return el.id === Number(id) });
+    listCopy[itemIndex].done = !listCopy[itemIndex].done;
+    this.setState({
+      list: listCopy
+    });
+  }
+  descriptionHandler = (newDescription) => {
+    const state = this.state;
+    state['newDescription'] = newDescription;
+    this.setState(state)
```
keepitterron: One comment was lost, sorry.
You are mutating the state in these functions, and you should avoid it at all costs. [here you can find out more why](https://medium.com/pro-react/a-brief-talk-about-immutability-and-react-s-helpers-70919ab8ae7c) you should keep the state immutable.

In this case you can simply do:
```js
descriptionHandler = newDescription => this.setState({ newDescription });
```

homework-submission/reactweek1/src/App.js
amirno1: you are right, i modified it the way that it is not mutating the state. Would be good if you take a look at it and let me know how you think about it.

homework-submission/reactweek1/src/App.js
amirno1: I also changed this one with filter..

homework-submission/reactweek1/src/App.js
amirno1: wow, fency stuff  :) was a good one thanks.

homework-submission/reactweek1/src/App.js
```diff
@@ -0,0 +1,152 @@
...
+
+  addHandler = () => {
+    const state = this.state;
+    const newItem = {
+      id: this.state.list.length + 1,
+      description: this.state.newDescription,
+      deadline: this.state.newDeadline,
+      done: false
+    };
+    ([...this.state.newDeadline] = ""), ([...this.state.newDescription] = "");
```
keepitterron: I'm not sure what this code is supposed to do, it does nothing at the moment.
I guess you want to reset those 2 state properties, why not adding them to the `setState` below?
```js
this.setState({
  list: [...state.list, newItem],
  newDeadline: '',
  newDescription: ''
});
```

homework-submission/reactweek1/src/App.js
```diff
@@ -0,0 +1,152 @@
...
+
+    this.setState({ list: [...state.list, newItem] });
+  };
+
+  removeHandler = i => {
+    let state = this.state;
+    const updatedList = state.list.filter(prop => prop.id !== i + 1);
```
keepitterron: why `i + 1`? this seems totally random :)
why don't you pass the id to this function instead of passing an index?

homework-submission/reactweek1/src/App.js
```diff
@@ -0,0 +1,152 @@
...
+            onChange={e => this.descriptionHandler(e.target.value)}
+            value={this.state.list.newDescription}
+          />
+        </label>
+        <fieldset>
+          <legend>Choose your date</legend>
+          <div>
+            <input
+              type="date"
+              min="2018-01-01"
```
keepitterron: 🎉

homework-submission/reactweek1/src/App.js
amirno1: ok there you go.

homework-submission/reactweek1/src/App.js
amirno1: I do not know, i just came up with that idea but your idea is indeed better.

amirno1:
@keepitterron Thanks , i really appriciate your constractive feedbacks.

PR #229, opened 2018-09-12
Neveenatik week4

neveenatik:
function handleEdit is not working. I'm not sure how to pass the id.
my code is messy. I am not sure I'm using the store right


homework-submission/week4/src/stores/TodoList.js
```diff
@@ -0,0 +1,100 @@
...
+  removeTodo = (id) => {
+    id = parseInt(id, 10);
+    return this.items = this.items.filter(item => item.id !== id);
+  };
+
+  @action 
+  editTodo = (id) => {
+    id = parseInt(id, 10);
+    return this.items = this.items.map((item) =>
+      item[id] === id
```
NouranMahmoud: Here, you are accessing the object property using the `item[id]` way but this means you are accessing `item.1` is this possible? in this case you need to access the id property like this `item.id`

neveenatik: Yes of course. I changed the way I access id but I didn't pay attention to update that and I kept reading it as id. 
Thank you for your time and review 🙏

PR #243, opened 2018-10-17
class16-week1-homework

AllouhM:


homework-submission/class16-week1/src/components/Todolist.js
```diff
@@ -0,0 +1,34 @@
+import React, { Component } from "react";
+import Todo from "./Todo";
+
+class Todolist extends Component {
+  render() {
+    const list = [
```
NouranMahmoud: It is better to take this const out of the `render` function. But it before the class or in a standalone file.

homework-submission/class16-week1/src/components/Todolist.js
```diff
@@ -0,0 +1,34 @@
...
+    const activity = list.map((item, i) => {
+      return (
+        <Todo key={i} description={item.description} deadline={item.deadline} />
+      );
+    });
+
+    return (
+      <div>
+        <h1 className="todos_header">My Todolist</h1>
+        <ul id="todo_list">{activity}</ul>
```
NouranMahmoud: **This is CSS related comment**

Usually, it is not preferable to use `id`, use class instead. classes gives more flexibility to overwrite styles in scalable projects in general.

Read about [CSS selectors specificity](https://css-tricks.com/specifics-on-css-specificity/).

homework-submission/class16-week1/src/components/Todo.js
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+
+class Todo extends React.Component {
+  render() {
+    const { description, deadline } = this.props;
+    return (
+      <li className="todo_item">
+        <h2>
+          Description{" "}
```
NouranMahmoud: What is the point from having `{" "}` ? adding a space? this should be accomplished by CSS. add `margin-right` to `.todo_desc`

homework-submission/class16-week1/src/components/Todo.js
```diff
@@ -0,0 +1,24 @@
...
+          <span className="todo_desc">
+            <em>{description}</em>
+          </span>
+        </h2>
+        <h3>
+          Deadline{" "}
+          <span className="todo_deadline">
+            <em>{deadline}</em>
```
NouranMahmoud: Instead of using `<em>` you can make the same effect with CSS by `font-style: italic` [check this example](https://www.w3schools.com/cssref/tryit.asp?filename=trycss_font-style)

homework-submission/class16-week1/src/components/Todo.js
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+
+class Todo extends React.Component {
+  render() {
+    const { description, deadline } = this.props;
+    return (
+      <li className="todo_item">
+        <h2>
```
NouranMahmoud: We usually use `<h2>` and `<h3>` if we have **Headers** but in this case this is normal text. So keep it simple

```
<li className="todo_item">
  <span className="desc_label">Description</span>
  <span className="desc_value">{description}</span>
</li>
```

In css you can add whatever style you like using the classes.

homework-submission/class16-week1/src/components/Todolist.js
AllouhM: I moved the list const out of the class.

homework-submission/class16-week1/src/components/Todolist.js
AllouhM: Yes, I know, but here as long as I have only one ul so I tended to make it unique. however, I changed it.

homework-submission/class16-week1/src/components/Todo.js
AllouhM: This has been added by default as a result of using Prettier plugin in VS.

homework-submission/class16-week1/src/components/Todo.js
AllouhM: This has been changed as well

homework-submission/class16-week1/src/components/Todo.js
AllouhM: this has been amended as well.

PR #245, opened 2018-10-18
week1 homework

sima-milli:


homework-submission/week1/src/todo.js
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+
+class Item extends React.Component {
```
NouranMahmoud: Currently that's fine, but It is better to split up your components in different files for scalability.

homework-submission/week1/src/todo.js
```diff
@@ -0,0 +1,28 @@
...
+    render() {
+        const { text, deadLine } = this.props;
+        return (
+            <li key={deadLine}>{text}, {deadLine}</li>
```
NouranMahmoud: Currently this is okay also, but the key should be unique like an **id**.

homework-submission/week1/src/index.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import './index.css';
+// import App from './App';
+import * as serviceWorker from './serviceWorker';
+import TodoList from "./todo";
+
+ReactDOM.render(<TodoList />, document.getElementById('root'));
```
NouranMahmoud: It's better to stick to "<App />" and within the App.js you can add whatever component you like.

The reason behind this is, There should be a single main "parent" component for all of your upcoming new features, also you would use App component for adding Router and Routes in future classes.

PR #246, opened 2018-10-18
week1 hw-1.0

mehmetscoskun:


SarKurd:
The app works and looks nice but there are a few things needs to be changed.

firstly the requirement was using two `components` but you have used only one. No, `change-format-date.js` isn't a component but a plain vanilla Javascript function.

You could make awrapper component and call it `TodoList.js` and render todos inside it then import `TodoList.js` in `App.js`

homework-submission/week1-react/src/App.js
```diff
@@ -0,0 +1,26 @@
...
+  render() {
+    return (
+      <div className="App">
+        <header className="App-header">
+
+          <img src={todoLogo} className="todo-logo" alt="todo" />
+          <div class="text-box">
+            <h2>Todo List</h2>
+
+            <Todo class="todo-item" item="go to Noordwijkerhout" date='2018-10-17' />
```
SarKurd: In React we use `className` instead of `class` to avoid collision with `class` keyword in javascript.
Also try to style components in their respective css file, todo.js can have a css file and styled in its css file

homework-submission/week1-react/src/change-format-date.js
```diff
@@ -0,0 +1,14 @@
+function changeDateFormat(strDate) {
```
SarKurd: As i said before too, it's not a react component

homework-submission/week1-react/src/App.js
SarKurd: Also it will be better to give the props a better name, for example `description` and `deadline`

homework-submission/week1-react/src/change-format-date.js
SarKurd: It's good that you have a utility file

PR #247, opened 2018-10-19
create props and test state

WardMad:


SarKurd:
Your app works but there are some unnecessary stuff that do nothing or not required

The `datepicker` and the state basically does nothing but it would be needed for future homeworks when adding a new todos, so it's great you know how to use it

The paragraph and logo and anything related to them are not required in this homework so i'm gonna skip them for during review

homework-submission/todo_react/src/task.js
```diff
@@ -0,0 +1,30 @@
...
+        deadline: new Date(),
+    }
+
+    onChange = deadline => this.setState({ deadline })
+
+    render() {
+        return (
+            <div>
+                <p>The task: {this.props.description}, todo till: {this.props.deadline} </p>
+                <p>{this.props.children}</p>
```
SarKurd: Line 17 does notthing since you have no children in App.js for the parent

WardMad:
Thanks for feedbacks:-)

PR #249, opened 2018-10-20
Week1 homework

marwanaboudaoud:


SarKurd:
The requirement was to use two components but i see only one

- you could make two components, one of them TodoList.js which wraps Todo.js and call it in app.js

- Make the files more organized by having a folder for components

homework-submission/my-project/src/page-footer.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import Picture from './all-done.jpg'
+
+class footer extends React.Component {
+    render() {
+        return <footer>
+            <div>
+                <img class="done" src={Picture} alt='allDone-logo' />
```
SarKurd: We use `className` instead of `class` in React

homework-submission/my-project/src/page-header.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import Picture from './todolist_logo.gif'
+
+class header extends React.Component {
+    render() {
+        return <header>
+            <div>
+                <img class="logo" src={Picture} alt='todo-logo' />
```
SarKurd: Same here we use `className`

homework-submission/my-project/src/todo.js
```diff
@@ -0,0 +1,21 @@
...
+
+const todos = [
+    { id: 1, description: 'Do my dutch homework', deadline: 'FRI 19 SEP' },
+    { id: 2, description: 'React homework', deadline: 'SAT 20 SEP' },
+    { id: 3, description: 'preaper to React week2 ', deadline: 'SUN 21 SEP' }
+
+]
+
+class TodosList extends React.Component {
+    item = todos.map((item) => (
```
SarKurd: `item` has no data type

homework-submission/my-project/src/todo.js
```diff
@@ -0,0 +1,21 @@
...
+        <li >
+            {item.id + "- " + item.description + " DeadLine: " + item.deadline}
```
SarKurd: - You could have another component for each `li`
- the list is already odered list so we don't need `item.id`, laso don't use item.id for indexing but the index which map return. here you called it `i`

- You could use template literal here

homework-submission/my-project/src/todo.js
```diff
@@ -0,0 +1,21 @@
...
+        </li>
+    ));
+    render() {
+        return <ol>{this.item}</ol>;
```
SarKurd: just use `item` alone without `this` and give it a data type

PR #250, opened 2018-10-20
Barbaros Week1

whitepiolin:


SarKurd:
+1 for the styling 

- Same as the other students the requirement was to use two components but i see only one. you could make another component and name it TodoList.js which wraps `staticToDo.js` and call it in app.js

- Make the files more organized by having a folder for components and give the file better names, for example `staticToDo.js` should be called `Todo.js` or `Item.js`

homework-submission/barbarosapp/src/App.js
```diff
@@ -0,0 +1,36 @@
...
+          <Item
+            name="Renovation of House"
+            deadLine="01/12/2018"
+            check="Undone" />
+        </section>
+      </div>
+    );
+  }
+}
+
```
SarKurd: You could save the data in an Array of objects or a json file and use map on it to make it more dynamic, what would you do if you had more than 10 tasks?

Also it makes sense to use a list here

whitepiolin:
Hi @SarKurd . 
Thanks for your nice comments. I have corrected my code by adding one more component and JSON file. Would you please check the new arrangements?
Thanks again :)

SarKurd:
Looks much better but still you are not dynamiclly rendering the todos in `ItemList.js`
use map to iterate over all of them and render them

You can do this
``` javascript

class ItemList extends React.Component {

  render() {
    const todos = data.map(todo => (
      <section className="ToDoItem">
      <Item
        name={todo.name}
        deadLine={todo.deadline}
        check={todo.situation} />
    </section>));

    return (
      <div className="ToDoList">
        {todos}
      </div>
    );
  }
};
```

homework-submission/barbarosapp/src/components/itemList.js
```diff
@@ -1,5 +1,7 @@
 import React from "react";
-import Item from './item'
+import Item from './item';
+
+let data = require('./data/data.json')
```
SarKurd: Use ES6 

`import data from './data/data.json'`

homework-submission/barbarosapp/src/components/itemList.js
```diff
@@ -0,0 +1,36 @@
+import React from "react";
+import Item from './item'
+
+class ItemList extends React.Component {
+
+  render() {
+    const { name, deadLine, check } = this.props;
```
SarKurd: Not used so delete it

whitepiolin:
Hi @SarKurd. I used more dynamic code in order to create the to do list as you helped. Also i corrected the data sourse according to ES6.
Thanks

PR #252, opened 2018-10-21
Shadi's homework week2

shadoo77:


SarKurd:
You app has no checkboxs and clicking on the todos doesn't change anything except color
You may want to give it a another try and make it work

SarKurd:
+1 for styling, using json file and file organizing

We would appreciate if you do what was asked in the home work, the whole adding item and deleting them are for the future homeworks and we have to review your code again,.

There are few stuff needs to be improved or fixed

- The checkbox are basically a visual change and it doesn't change the actual global state.
- Using the local states for each todo is a wrong idea and since you already have a global state and made the code complex and unreadable. Anything you have done could be done with the global state only
- functions, variables and folders should have better names

 The code overall need refactoring and follow above feedbacks

homework-submission/todo-app/src/todos/addform.jsx
```diff
@@ -0,0 +1,101 @@
+import React, { Component } from 'react';
+import DatePicker from 'react-datepicker';
+import moment from 'moment';
+ 
+import 'react-datepicker/dist/react-datepicker.css';
+
+class SendForm extends Component {
```
SarKurd: That's not how we do form validation in react, use the state to listen to the changes and check the state then based on that changes send a response to the user

homework-submission/todo-app/src/todos/element.jsx
```diff
@@ -0,0 +1,42 @@
+import React, { Component } from 'react';
+
+class Element extends Component {
+    state = {
+        id: this.props.id,
+        description: this.props.description,
```
SarKurd: Local state not necessary

homework-submission/todo-app/src/todos/element.jsx
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    checkItemsDone() {
+        const {description, deadline} = this.state;
+        const item = description + ', ' + deadline;
+        return this.state.done 
+            ? <p onClick={this.throwLineHandle}><del>{item}</del></p> 
+            : <p onClick={this.throwLineHandle}>{item}</p>
+    }
+    myCheckBox() {
```
SarKurd: You could make a checkbox component to make more reusable and easy to read

homework-submission/todo-app/src/todos/todos.jsx
```diff
@@ -0,0 +1,69 @@
...
+import Data from './data';
+import AddForm from './addform';
+
+
+class Todos extends Component {
+    constructor(props) {
+        super(props);
+        this.state = { 
+            elements: Data
+         }
```
SarKurd: that's the only state that should be modified and send this state as props

homework-submission/todo-app/src/todos/todos.jsx
```diff
@@ -0,0 +1,69 @@
...
+        const arrSize = this.state.elements.length;
+        const newItem = {
+            id: arrSize + 1,
+            description: formFields.description,
+            deadline: formFields.deadline,
+            done: formFields.done
+        };
+        const elements = this.state.elements;     
+        elements.push(newItem);
+        this.setState({elements});
```
SarKurd: I can say you understood how to mutate the state, +1

homework-submission/todo-app/src/todos/todos.jsx
```diff
@@ -0,0 +1,69 @@
...
+    listOfElements() {
+         return <ul className="list-group list-group-hover list-group-striped">
+                    { this.extractElements() }
+                </ul>;
+    }
+
+    checkElementsMount() {
+        const noElement =  <div className="alert alert-danger" role="alert">No items...!</div>;
+        return this.state.elements.length < 1 ? noElement : this.listOfElements();
+    }
```
SarKurd: all these three function could have been one simple function

shadoo77:
Oeps!
It has wel checkbox and locally it works with me!!
It changes the done value

Op wo 31 okt. 2018 om 00:41 schreef Sarbast Mohammed Mustafa <
notifications@github.com>

homework-submission/todo-app/src/todos/addform.jsx
shadoo77: here i didn't want to use any validation ! 
and the checkbox works as expected ! weird , i'm behind my laptop , everything works!

homework-submission/todo-app/src/todos/element.jsx
shadoo77: description is necessary , maybe id not okey!

homework-submission/todo-app/src/todos/element.jsx
shadoo77: yeah i'm working on it now

homework-submission/todo-app/src/todos/todos.jsx
shadoo77: here i didn't understand what you mean , can you give me some explanation ?

SarKurd: Javascript objects and arrays are copied by reference so when you are changing the copy element, it will change the original object too

to make a  copy so the original object doesn't change when changing the copy object, we have to use shallow copy, slice() or spread operator which make a shallow copy to the top elements in the objects(don't work for nested objects)

safest way to make a copy so the nested objects doesn't get copied by reference is to use deep copy, a for loop and iterate over all items or shallow copy for each object and nested object or json stringify and json parse(but doen't work if u have function inside the  object)

https://www.youtube.com/watch?v=EeZBKv34mm4&ab_channel=freeCodeCamp.org

PR #253, opened 2018-10-23
week2 homework

Dumie1:


homework-submission/myapp/src/Components/Data.js
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    renderTodos() {
+        if (this.state.Todos.length === 0) return <p className='noItems'>No items...</p>
+
+        const textStyle = { textDecoration: 'line-through', backgroundColor: '#0455d8' }
+
+        return <ul>
+            {
+                Todos.map((todo, id) => {
```
SarKurd: line 24, you have to use the data inside the state here
`this.state.data`

homework-submission/myapp/src/Components/Data.js
```diff
@@ -0,0 +1,47 @@
...
+        todo.done = !todo.done;
+        this.setState({
+            Todos
+        });
+    }
+
+    renderTodos() {
+        if (this.state.Todos.length === 0) return <p className='noItems'>No items...</p>
+
+        const textStyle = { textDecoration: 'line-through', backgroundColor: '#0455d8' }
```
SarKurd: no need to write a function, the contetnt of this function should be inside the render function

homework-submission/myapp/src/Components/Data.js
```diff
@@ -0,0 +1,47 @@
...
+
+        return <ul>
+            {
+                Todos.map((todo, id) => {
+
+                    return (
+                        <li key={id}>
+                            <input checked={todo.done} type='checkbox' onChange={this.isTaskDone.bind(this, todo)} />
+                            <span style={todo.done ? textStyle : null}>
```
SarKurd: line 28, you don't need to bind since you are already using arrow functions

try this: 
```javascript
onChange={() => this.isTaskDone(todo)}
````

File: homework-submission/myapp/src/Components/Data.js [hunk 4]
```diff
@@ -0,0 +1,47 @@
...
+import Todos from './Todos.json';
+
+class Data extends React.Component {
+
+    state = {
+        Todos
+    }
+
+    isTaskDone = (todo) => {
+        todo.done = !todo.done;
```
SarKurd: don't mutate the state directly IMPORTANT, you are mutating the state then using setState !!! it doen't make sense
make a deep copy and mutate then use set state

homework-submission/myapp/src/Components/Data.js
Dumie1: I have deleted renderTodos function and moved the above code inside the render method.

homework-submission/myapp/src/Components/Data.js
Dumie1: Ohh!! my mistake. I have sent you a message on slack about this part... Thank you for the constructive feedback.

PR #254, opened 2018-10-23
Week2

genegr88:
Finished homework met all requirements of homework but could not fix the problem with AddForm component


SarKurd:
+1 for a clean and working great app

Here the things to be changes
- The requirement was using 2 components but there is only one. Move the state and functions to a separate component and call it `TodoList.js`, now you will have 2 components. Always try to make `App.js` as clean as possible
- AddForm component is there but not used
- No file organizing except for the data
- Almost no styling

homework-submission/homework-week1/src/App.js
```diff
@@ -0,0 +1,43 @@
...
+    this.setState({
+      todos
+    })
+  }
+
+  checkBoxHandler = (id) => {
+    const todos = this.state.todos.map(item => {
+      if (item.id === id) {
+        return {
+          ...item, done: !item.done
```
SarKurd: +1 for having two clean functions and they work great(deleteTodo and checkboxHandler)

PR #260, opened 2018-10-27
week2 homework

sima-milli:


SarKurd:
+1 for following the requirements and works great

There are a few things needs to be changed
- state and the handle function should be inside `todoList.js`
- you are mutating the state directly, don't do that

homework-submission/week2/src/App.js
```diff
@@ -0,0 +1,32 @@
...
+import todos from './todo/todos';
+
+class App extends Component {
+  state = {
+    todos,
+  };
+
+  handelChange = id => {
+    const newTodos = this.state.todos;
+    newTodos.map(todo => {
```
SarKurd: line 12: don't mutate the state directly make a deep copy, important

sima-milli: Thanks @SarKurd for your review 👍

PR #263, opened 2018-10-28
Sam week2

Samtabanja:
done


SarKurd:
where is the package.json file? i will review your code after pushing it

PR #264, opened 2018-10-28
homework week2

ozgurleo:


todoapp/src/App.js
```diff
@@ -0,0 +1,105 @@
...
+    })
+  }
+
+  getNumberOfReadtodos(todos) {
+    return todos.filter(comment => !comment.read).length;
+  }
+
+  submitNewItem(event) {
+    event.preventDefault();
+    const updatednewToDoItem = Object.assign({}, this.state.newToDoItem, {
```
Jawhar-B: Like the previous comment about `Object.assign`
```suggestion
    const updatednewToDoItem = {...this.state.newToDoItem, 
      date: new Date(),
      id: this.state.todos.length + 1,
      read: false
    }
```

todoapp/src/App.js
```diff
@@ -0,0 +1,105 @@
...
+      date: new Date(),
+      id: this.state.todos.length + 1,
```
Jawhar-B: What is the `id` for the next one of this if this `todo` has been deleted?

todoapp/src/App.js
```diff
@@ -0,0 +1,105 @@
...
+    const updatedtodos = [...this.state.todos, updatednewToDoItem];
+    this.setState({
+      todos: updatedtodos,
+      newToDoItem: {
+      text: ''
+      }
+    });
+  }
+
+  inputAdd(event) {
```
Jawhar-B: we may have another name for this method because of the naming importance:
so you have `inputAdd` and for me first I read it I thought it's for submitting which is not.

you may call it something like `handleChange` or similar it's very common thing in the React community to call it with these kind of names.

todoapp/src/App.js
```diff
@@ -0,0 +1,105 @@
...
+      newToDoItem: updatednewToDoItem
+    });
+  }
+
+  render() {
+    return (
+      <div>
+        <h1>
+          ToDo List {
+            this.state.todos.length !== 0 && `(unread ${this.getNumberOfReadtodos(this.state.todos)})`
```
Jawhar-B: while you are using a method from the component it self it's enough to say:
```suggestion
            this.state.todos.length !== 0 && `(unread ${this.getNumberOfReadtodos()})`
```

because it's very important make a note you are taking from the state in the function it self, By make the use of state there, unless you want to use the function other ware as well 🙂

todoapp/src/App.js
```diff
@@ -0,0 +1,105 @@
...
+        return Object.assign({}, comment, {read: updatedReadValue});
+      }
+      return comment;
+    });
+    this.setState({
+      todos: updatedtodos
+    });
+  }
+
+  rendertodos(todos) {
```
Jawhar-B: `rendertodos`
**camelCase 😬**

the other thing is not required but I think it's better to have a separated file for a new component you are creating.

it's good use of a very tiny change in a methods, but if you used to have more than a one file you will have better understanding of each component on it's own.

you can pass any needed things here to it by `props` and it will work exactly the same.

so you don't have to import all your components in the App component file but making them on a multiple files.

todoapp/src/App.js
```diff
@@ -0,0 +1,105 @@
...
+            <textarea
+              name="text"
+              value={this.state.newToDoItem.text}
+              onChange={this.inputAdd}
+            />           
+            <button>submit</button>
+          </form>
+        </div>
+
+        {this.rendertodos(this.state.todos)}
```
Jawhar-B: if you have get my last comment on this as an advise you may have the following component...
```suggestion
        <Todos todos={this.state.todos} changeToDoList={this.changeToDoList}/>
```
and I changed the name to be more relevant with the React naming.

PR #268, opened 2018-11-01
week3 homework

Dumie1:


homework-submission/myapp/src/Components/Data.js
```diff
@@ -0,0 +1,93 @@
...
+  addTodo = event => {
+    event.preventDefault();
+
+    const descriptionInput = this.descriptionInput.current.value;
+    const deadlineInput = this.deadlineInput.current.value;
+    const newState = this.state;
+
+    const Todos = newState.Todos;
+    Todos.push({
+      id: Todos.length + 1,
```
Jawhar-B: What is the `id` for the next one of this if this `todo` has been deleted?

Well in your case you are re-rendering the whole thing, which is fine.

But in a different situation you will face a Bug may be hard to follow or fix, so may have to re-think about how do I specify an `Id` for a new Item.

homework-submission/myapp/src/Components/Data.js
```diff
@@ -0,0 +1,93 @@
...
+
+    this.setState(newState);
+
+    this.descriptionInput.current.value = '';
+    this.deadlineInput.current.value = '';
+  };
+
+  deleteTodo = id => {
+    const arr = this.state.Todos;
+    arr.splice(id, 1);
```
Jawhar-B: same as a previous comment, this is mutating the state immediately, perhaps `.filter` Method ?

Remember the checking test:
```javascript
    console.log(this.state.Todos)
    arr.splice(id, 1);
    console.log(this.state.Todos)
```

homework-submission/myapp/src/Components/Data.js
```diff
@@ -0,0 +1,93 @@
...
+    if (this.state.Todos.length === 0)
+      return <p className='noItems'>No items...</p>;
+    const textStyle = {
+      textDecoration: 'line-through',
+      backgroundColor: '#0455d8'
+    };
+
+    return (
+        <ul>
+        {Todos.map((todo, id) => {
```
Jawhar-B: where `Todos` come from? 🙃

Dumie1: Instead of `Todos.map` , my corrections are as follows `this.state.Todos.map` ... i hope im on the right track

homework-submission/myapp/src/Components/Data.js
Dumie1: Hahaha i get what you mean. Probably `.splice()` method was easy for me to use. However thanks @Jawhar-B for the constructive feedback. i will try using `.filter()` method.

PR #269, opened 2018-11-01
class16-week3

AllouhM:


AllouhM:
Thank you Jawhar for the useful feedback, I amended my code to reflect your notes 👍

PR #271, opened 2018-11-03
Barbaros Week3

whitepiolin:


homework-submission/barbarosapp/src/App.js
```diff
@@ -0,0 +1,62 @@
...
+      deadline,
+      done: false
+    };
+    this.setState({ Data: [...addedData, addition] });
+  };
+
+  handleRemove = item => {
+    const newData = this.state.Data;
+    newData.splice(item.id, 1);
+    this.setState({ Data: newData });
```
Jawhar-B: Two things are here:
 1. you are mutating the state directly with splice.
 2. you may use `.filter` instead of splice making better performance for react while `.filter` is returning a new array, it means you have a complete new state.

homework-submission/barbarosapp/src/components/item.js
```diff
@@ -0,0 +1,36 @@
...
+            Name : <span className="Color1"> {name} </span>
+          </p>
+          <p>
+            Deadline : <span className="Color2"> {deadLine} </span>
+          </p>
+        </div>
+
+        <div className="removeButtonDiv">
+          <button
+            onClick={item => this.handleRemove(item.id)}
```
Jawhar-B: Here you have `onClick` event, my question is what is the value of `item.id` here?

whitepiolin: Yes, I corrected remove button problem. Thanks

homework-submission/barbarosapp/src/components/item.js
```diff
@@ -0,0 +1,32 @@
+import React from "react";
+import CheckBox from "./checkBox";
+import RemoveButton from "./removeButton";
+
+class Item extends React.Component {
+  render() {
+    const { name, deadLine, handleCheck, handleRemove, done, id } = this.props;
+
```
Jawhar-B: I will complete the review after this hint. 
```suggestion
console.log(deadLine)
```

if you get stuck and you want me to continue let me know 😬

Jawhar-B:
By the way you have a very nice styling 😃👌

homework-submission/barbarosapp/src/components/item.js
whitepiolin: Corrected. Thanks for help

whitepiolin:
Thanks Jawhar-B for helping me when i was stucked. Now my code is running and in more efficient way.

homework-submission/barbarosapp/src/App.js
whitepiolin: Corrected by using filter

homework-submission/barbarosapp/src/App.js
```diff
@@ -0,0 +1,68 @@
...
+      }
+      return false; 
+    }
+
+    const deletedData = this.state.Data.filter(filterByID);
+    this.setState({ Data: deletedData });
+  };
+
+  handleAdd = (name, deadline) => {
+    const newId = this.state.Data.length;
```
Jawhar-B: This is a `bug`

ezgif com-video-to-gif 1

Ok let's just break it down!

let's assume that we have three elements:
 1. two are already there with `id`s 1, 2
 2. I removed the first one, which have `1` as an `id` and added a new one

So with that we will have the following `id`s [2, 3], next we added a new two elements:
> according to the new length we will have the following `id`s

[2, 3, 1, 2] if I added a new one I will fil in the blank the second `3` which is already there, don't you agree?

------
for solving this it's probably be better to stay away from the `.length` property, and maybe using another methodology that may solve too many things, like `uuid` or something other. 🙂

homework-submission/barbarosapp/src/App.js
```diff
@@ -0,0 +1,68 @@
...
+    const newData = this.state.Data;
```
Jawhar-B: Correction: this is not a new but it's the current state of data, just not making you confused with passing by `value` or `reference`.

homework-submission/barbarosapp/src/App.js
whitepiolin: Yes, I can create items with same id's. So i will try to use guid

homework-submission/barbarosapp/src/App.js
whitepiolin: Yes, corrected

PR #278, opened 2018-11-11
Week3

acimanx:


Jawhar-B:
Hi @acimanx, I am not sure about the way of reviewing your homework, you have an amazing ideas and you are always so encouraged to learn more and more which what I like.


I played your application it started with this:
screenshot 2018-11-22 at 11 28 19

I solved it to find this:
screenshot 2018-11-22 at 08 35 42

then this: 
screenshot 2018-11-22 at 11 30 53

Even if it was a warnings but it's not rendering:
screenshot 2018-11-22 at 11 31 56

Well until now it's still **okay**! So I didn't give up and solved it to finally it works 🎉😃

I am not going to review this application until it works properly at least, plus it have to be as the requirements of the homework

this might be annoying I am sorry about it but this harmed me!!!
https://stackoverflow.com/questions/35823783/finding-the-array-index-of-an-object-with-javascript-react-js

PR #282, opened 2018-11-15
week4 homework

Dumie1:


homework-submission/myapp4/src/App.js
```diff
@@ -0,0 +1,71 @@
...
+      return <p className="noItems">No items...</p>;
+
+    const textStyle = {
+      textDecoration: "line-through",
+      backgroundColor: "#0455d8"
+    };
+
+    return (
+      <ul>
+        {this.props.TodosStore.Todos.map((todo, id) => {
```
Jawhar-B: While you have an `id` in your `todo` object, then what is `id` here? can it be instead?

homework-submission/myapp4/src/stores/TodosStore.js
```diff
@@ -0,0 +1,43 @@
+import { observable, computed, action } from "mobx";
+
```
Jawhar-B: Of I forgot this you missed too 🙂 
```suggestion
configure({ enforceActions: 'observed' })
```

PR #283, opened 2018-11-16
Week4 homework

end0620:


homework-submission/week1-project/src/Todo app/TodoApp.js
```diff
@@ -0,0 +1,26 @@
+import {observable, computed, action} from 'mobx';
```
Jawhar-B: You missed This 👌
```suggestion
import {observable, computed, action, configure} from 'mobx';
configure({ enforceActions: 'observed' });
```

PR #285, opened 2018-11-17
Mobx week1

sima-milli:


homework-submission/mobx-week1/src/App.js
```diff
@@ -0,0 +1,67 @@
+import React, { Component } from "react";
+import "./Componants/index.css";
+import { inject, observer } from "mobx-react";
+import TodoList from "./Componants/TodoList";
+// import DatePicker from "react-datepicker";
+// import moment from "moment";
+
+// import "react-datepicker/dist/react-datepicker.css";
```
Jawhar-B: If no need for these lines you can remove them.

homework-submission/mobx-week1/src/App.js
```diff
@@ -0,0 +1,67 @@
...
+                name="description"
+                className="form-control mb-2 mr-sm-2"
+                id="inlineFormInputName2"
+                placeholder="description"
+                ref={input => (this.newDescription = input)}
+              />
+              {/* <DatePicker
+                selected={this.TodoStore.deadline}
+                onChange={this.handleChange}
+              /> */}
```
Jawhar-B: This one too.

homework-submission/mobx-week1/src/Componants/TodoList.js
```diff
@@ -0,0 +1,113 @@
...
+  state = {
+    clicked: false
+  };
+
+  onEdit = () => {
+    this.setState({
+      clicked: true
+    });
+  };
+  onCancel = () => {
```
Jawhar-B: Oh! you have here `onCanel` and before it line 11 `onEdit`, also you have on line 27 a `setState` for setting the state to be as cancel.

I think we should avoid repeating code for a bit and may be do the following?
```javascript
  toggleEdit = () => {
    this.setState({
      clicked: !this.state.clicked
    });
  }
```

Now this method can be used every where you have edit event or cancel or after submit 🎉

homework-submission/mobx-week1/src/Componants/TodoList.js
```diff
@@ -0,0 +1,113 @@
...
+    const description = this.newDescription.value;
+    const deadline = this.newDeadline.value;
+
+    this.props.TodoStore.updateTodo(id, description, deadline);
+    this.setState({
+      clicked: false
+    });
+  };
+
+  editAction = todo =>
```
Jawhar-B: This method is confused me a bit, about where it's start and where it finished 🙄

homework-submission/mobx-week1/src/Componants/TodoList.js
```diff
@@ -0,0 +1,113 @@
...
+        />
+        <button type="submit" className="btn btn-success">
+          Update
+        </button>
+        <button className="btn btn-warning" onClick={this.onCancel}>
+          Cancel
+        </button>
+      </form>
+    ) : (
+      <span className="todo-text" style={checkboxStyle(todo.done)}>
```
Jawhar-B: Styling is one of the most important things, But what **If** you wanted to change the styling from your stylesheet file in the future.
This won't change, right? 🙃💔

there are many ways of styling improvements in the react world, Have you a look:
https://codeburst.io/4-four-ways-to-style-react-components-ac6f323da822

I saw you are using this function for a multiple times which is great but I would prefer to leave the css to it's own part, up to you, what do you say?

homework-submission/mobx-week1/src/stores/TodoStore.js
```diff
@@ -0,0 +1,40 @@
+import { observable, action, computed } from "mobx";
+// import moment from "moment";
+
+class TodoStore {
+  @observable todos = [];
+  @observable editing = false;
+
+  @action addTodo = todo => {
+    if (todo.description.length !== 0 && todo.deadline.length !== 0)
```
Jawhar-B: May this be?
```suggestion
    if (todo.description.length && todo.deadline.length)
```

homework-submission/mobx-week1/src/stores/TodoStore.js
```diff
@@ -0,0 +1,40 @@
...
+      this.todos.push(todo);
+  };
+  @action onCheck = (e, id) => {
+    this.todos = this.todos.map(todo => {
```
Jawhar-B: I don't see any use of `.map` here more than the iteration part, while we are looping through all the elements and mutating by there value, we are mutating them directly and not by reference.

with that using `.forEach` is a bit more reasonable, same as the next method.

homework-submission/mobx-week1/src/stores/TodoStore.js
```diff
@@ -0,0 +1,40 @@
+import { observable, action, computed } from "mobx";
```
Jawhar-B: You missed This 👌
```suggestion
import { observable, action, computed, configure } from "mobx";
configure({ enforceActions: 'observed' });
```

PR #286, opened 2018-11-17
Donal's hw4

Donal86:


homework-submission/todo-app/src/App.js
```diff
@@ -0,0 +1,30 @@
+import React, { Component } from 'react';
+import { inject, observer } from 'mobx-react';
+import Form from './Components/Form';
+import List from './Components/List';
+import './App.css';
+
+@inject('TasksStore')
+@observer
```
Jawhar-B: I would prefer to inject in the Form component as a one way code style.

because you are injecting here and in the `List` component, while you don't have a use of the store actions you are passing, but instead you are injecting. 💔

homework-submission/todo-app/src/App.js
```diff
@@ -0,0 +1,30 @@
...
+  render() {
+    const { TasksStore } = this.props;
+    return (
+      <div className="App">
+        <Form
+          onAdd={TasksStore.handleAdd}
+
+        />
+        <List
+          handleCheck={TasksStore.handleCheck}
```
Jawhar-B: As I said in the previous comment, None of these props are in use inside of the `List` Component.

homework-submission/todo-app/src/Components/Form.js
```diff
@@ -0,0 +1,54 @@
...
+import DatePicker from 'react-date-picker';
+
+
+class Form extends Component {
+    state = {
+        description: '',
+        deadline: new Date()
+    };
+
+    handleChange = evt => {
```
Jawhar-B: I would go for another good descriptive name such as `handleDescriptionChange`?

homework-submission/todo-app/src/Components/List.js
```diff
@@ -0,0 +1,37 @@
...
+@inject('TasksStore')
+@observer
+
+class List extends Component {
+    render() {
+        const { TasksStore } = this.props;
+        return (
+            <div className='List'>
+                <ul>
+                    {TasksStore.todos.map((todo, id) => (
```
Jawhar-B: What is the value of `id`?
`id` is for identifying, while index for what it means, don't we want to identify ? if yes 😃 then we may use `todo.id` instead of `id` ?

homework-submission/todo-app/src/Components/Task.js
```diff
@@ -0,0 +1,21 @@
+import React, { Component } from 'react';
+import '../App.css';
+
+class Task extends Component {
+    render() {
+        const { todo } = this.props;
+        return (
+            <div className='Task'>
+                <div style={{ textDecoration: todo.done ? 'line-through' : 'none' }}>
```
Jawhar-B: Styling is one of the most important things, But what **If** you wanted to change the styling from your stylesheet file in the future.
This won't change, right? 🙃💔

there are many ways of styling improvements in the react world, Have you a look:
https://codeburst.io/4-four-ways-to-style-react-components-ac6f323da822

homework-submission/todo-app/src/stores/TasksStore.js
```diff
@@ -0,0 +1,42 @@
+import { observable, action } from "mobx";
+import Data from "../data/todos.json";
+import uuid from 'uuid';
+
+class TasksStore {
+    @observable todos = Data;
+
+    @action
+    handleCheck = id => {
+        const newTodos = this.todos.map(todo => {
```
Jawhar-B: I don't see any use of `.map` here more than the iteration part, while we are looping through all the elements and mutating by there value, we are mutating them directly and not by reference.

with that using `.forEach` is a bit more reasonable.

homework-submission/todo-app/src/stores/TasksStore.js
```diff
@@ -0,0 +1,42 @@
+import { observable, action } from "mobx";
```
Jawhar-B: You missed This 👌
```javascript
import { observable, action, configure } from "mobx";
import Data from "../data/todos.json";
import uuid from 'uuid';

```

PR #287, opened 2018-11-17
Week4

whitepiolin:
Still working on it


whitepiolin:
Attention to the Teacher;
I have problems in checkbox. I couldn't effect the data of API by handleCheckBox function.

homework-submission/myapp/src/components/checkBox.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import { observer, inject } from "mobx-react";
+
+@inject("todoStore")
+@observer
+class CheckBox extends React.Component {
+  
+  render() {
+    const { todoStore, done, id } = this.props;
+
```
Jawhar-B: You didn't pass `done, id` while you are using them here
```suggestion
   console.log(done, id);
```

homework-submission/myapp/src/components/item.js
```diff
@@ -0,0 +1,39 @@
...
+@inject("todoStore")
+@observer
+class Item extends React.Component {
+  render() {
+    const { name, deadLine, id } = this.props;
+
+    return (
+      <section className="ToDos">
+        <div className="CheckBoxDiv">
+          <CheckBox />
```
Jawhar-B: - The Done Bug
```suggestion
          <CheckBox done={this.props.done} id={id} />
```

homework-submission/myapp/src/App.js
```diff
@@ -0,0 +1,24 @@
+import React, { Component } from "react";
+import "./App.css";
+import logo from "./logo.svg";
+import ItemList from "./components/itemList";
+import AddMenu from "./components/addMenu";
+import { observer, inject } from "mobx-react";
+
+@inject("todoStore")
+@observer
```
Jawhar-B: I would prefer to not inject in a component if no need to extract the store from it in case of use.

homework-submission/myapp/src/App.js
```diff
@@ -0,0 +1,24 @@
...
+class App extends Component {
+  render() {
+    return (
+      <div className="App">
+        <header className="App-header">
+          <img src={logo} className="App-logo" alt="logo" />
+        </header>
+        <AddMenu Data={this.Data} value={this.value} />
```
Jawhar-B: `this.Data` and `this.value` 🤔

homework-submission/myapp/src/components/addMenu.js
```diff
@@ -0,0 +1,81 @@
...
+  handleChangeDate = date => {
+    this.setState({ deadline: date });
+  };
+
+  handleSubmitAdd = event => {
+    event.preventDefault();
+    this.props.todoStore.handleAdd(this.state.name, this.state.deadline);
+    this.setState({
+      name: "",
+      deadline: moment()
```
Jawhar-B: Here you are setting it to be a `moment` while you have it by default as a string in your state.

I got this warning while starting the program:
screenshot 2018-11-29 at 02 40 31

with these both if no need for moment.js It's for a better performance.
By checking the Docs of [react-datepicker](https://www.npmjs.com/package/react-datepicker)

you will find out they [started using native js Date](https://www.npmjs.com/package/react-datepicker#momentjs) and you have the newest version there 😃

homework-submission/myapp/src/components/itemList.js
```diff
@@ -0,0 +1,35 @@
...
+import Item from "./item";
+import { observer, inject } from "mobx-react";
+
+@inject("todoStore")
+@observer
+class ItemList extends React.Component {
+  render() {
+    const { todoStore } = this.props;
+
+    const createItems = todoStore.Data.map((item, index) => {
```
Jawhar-B: `id` is for identifying used to be, while index for what it means, don't we want to identify ? if yes 😃 then we may use `item.id` instead of `id` as a key?

homework-submission/myapp/src/stores/todoStore.js
```diff
@@ -0,0 +1,49 @@
+import { observable, action } from "mobx";
+import Data from "./../data/data";
+
+class todoStore {
+  @observable Data = Data;
+
+  @action handleCheckBox = id => {
+    this.Data = this.Data.map(item => {
```
Jawhar-B: I don't see any use of `.map` here more than the iteration part, while we are looping through all the elements and mutating by there value, we are mutating them directly and not by reference.

with that using `.forEach` is a bit more reasonable, same as the next method.

homework-submission/myapp/src/stores/todoStore.js
```diff
@@ -0,0 +1,49 @@
...
+      if (item.id === id) {
+        return { ...item, done: !item.done };
+      }
+      return item;
+    });
+  };
+
+  @action handleRemove = id => {
+    function filterByID(item) {
```
Jawhar-B: 🎉
For more improvement this can be a bit shorter:
```javascript
    const filterByID = item => item.id !== id
```

homework-submission/myapp/src/stores/todoStore.js
```diff
@@ -0,0 +1,49 @@
...
+      }
+      return false;
+    }
+
+    const deletedData = this.Data.filter(filterByID);
+    this.Data = deletedData;
+  };
+
+  @action handleAdd = (name, deadline) => {
+    const uniqueId = () => {
```
Jawhar-B: Great, I liked this function 🎉🙌🏻

homework-submission/myapp/src/stores/todoStore.js
```diff
@@ -0,0 +1,49 @@
+import { observable, action } from "mobx";
+import Data from "./../data/data";
+
```
Jawhar-B: Oh! you missed a thing in your homework 👌
```suggestion
configure({ enforceActions: 'observed' })
```

whitepiolin:
Thanks for your helpful comments. I am a bit late about the corrections of React but this week i am moving a new house and this is really like in hell. I need to handle also project, So i will handle these corrections this week. Thanks again

PR #291, opened 2018-11-22
Week4

mehmetscoskun:


homework-submission/week1-react/src/App.js
```diff
@@ -0,0 +1,67 @@
...
+class App extends Component {
+
+  render() {
+    const {myTodoApp} = this.props;
+
+    const isEmpty = myTodoApp.data.todos.length < 1 || myTodoApp.data.todos === undefined;
+
+    const todoList = myTodoApp.data.todos.map((item, key)=> 
+    (key === myTodoApp.data.indexToUpdate) ?
+    <Update
```
Jawhar-B: This is returning a `form` which is not a valid child for `ul` element.

homework-submission/week1-react/src/App.js
```diff
@@ -0,0 +1,67 @@
...
+@observer
+
+class App extends Component {
+
+  render() {
+    const {myTodoApp} = this.props;
+
+    const isEmpty = myTodoApp.data.todos.length < 1 || myTodoApp.data.todos === undefined;
+
+    const todoList = myTodoApp.data.todos.map((item, key)=>
```
Jawhar-B: What is the value of `key` here?
`id` is for identifying, while index for what it means, don't we want to identify ? if yes 😃 then we may use `item.id` instead of `key` here?

homework-submission/week1-react/src/App.js
```diff
@@ -0,0 +1,67 @@
...
+
+@inject('myTodoApp')
+@observer
+
+class App extends Component {
+
+  render() {
+    const {myTodoApp} = this.props;
+
+    const isEmpty = myTodoApp.data.todos.length < 1 || myTodoApp.data.todos === undefined;
```
Jawhar-B: This may be shortened a bit for a code styling if you would like 🙂:
```suggestion
    const isEmpty = !myTodoApp.data.todos.length || !myTodoApp.data.todos;
```
Or may be :
```suggestion
    const isEmpty = !(myTodoApp.data.todos.length || myTodoApp.data.todos);
```

homework-submission/week1-react/src/component/update.js
```diff
@@ -0,0 +1,47 @@
...
+      })
+    }
+
+    handleDeadline (event) {
+        this.setState({
+            deadline: event.target.value
+        })
+    }
+
+    handleSubmit (event) {
```
Jawhar-B: Please look this up: https://github.com/HackYourFuture/React/pull/274#discussion_r235436549

homework-submission/week1-react/src/component/todo.js
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+import changeDateFormat from '../util/change-format-date.js';
+
+class Todo extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
```
Jawhar-B: I see you have forgotten this in here while you are not using it. 😅

homework-submission/week1-react/src/component/todo.js
```diff
@@ -0,0 +1,28 @@
...
+        };
+    }
+
+    render () {
+        const {text, date, done, changeDone, deleteTodo, index, editButton} = this.props;
+
+        return(
+                <li>
+                    <label className= {done ? 'textFormat' : null}>
+                        <input type="checkbox" className="checkbox" onChange={()=>changeDone(index)} defaultChecked={this.props.done}/>
```
Jawhar-B: Please look up: https://github.com/HackYourFuture/React/pull/274#discussion_r235435327

homework-submission/week1-react/src/component/todos.json
```diff
@@ -0,0 +1,20 @@
+[
```
Jawhar-B: In my opinion changing the path to be equivalent to the data it's inside well make a better understanding for the file structuring.

homework-submission/week1-react/src/App.js
mehmetscoskun: Yes, I made change and now it is like below:
```
<li>
      <form> 
                  ...
      <form/>
<li/>
```

homework-submission/week1-react/src/App.js
mehmetscoskun: Actually I added `key` because of I got an error. But now I removed everything I did not used. It looks prettier. I am using just `id={item._id}` for iterator and identification

homework-submission/week1-react/src/App.js
mehmetscoskun: Thank you for taking my attention to `truthy falsy` . actually I know what it is but I was not using them directly. it looks simple

homework-submission/week1-react/src/component/todo.js
mehmetscoskun: Yepp :)

homework-submission/week1-react/src/component/todo.js
mehmetscoskun: yepp, just `checked={done}` works, thanks

homework-submission/week1-react/src/component/update.js
mehmetscoskun: To understand what you advise, I took a look at stackoverflow. I see this:
"When you use the <button> element without specifying a type, it defaults to type="submit". You need to specify it as type="button" to stop it submitting the form:"
And then I removed `handleSubmit` function and changed `type="submit"` as `type="button"`  Now it works as previously. But I could not realize the difference

mehmetscoskun:
Thank you Jawhar for your valuable advices.  I hope no `needs some work` point left :) bye

PR #294, opened 2018-11-24
Barbaros Week5

whitepiolin:


whitepiolin:
@Jawhar-B Thanks for your helpful comments. I am a bit late about the corrections of React but this week i am moving a new house and this is really like in hell. I need to handle also project, So i will handle these corrections in a week. Thanks again

PR #301, opened 2019-01-08
week1 homework

kawaraa:


jalal75:
- Please read the homework carefully:
    ( 
        * You have to use more than two components.
        * Think which props the components should take.
    )
Do some style even if it is simple.
you don't need to push every file, just your own files.

PR #302, opened 2019-01-08
:boom: HW uploaded

moesi88:


moesi88:
firrst HW react

week1-HW/todo-app/src/App.js
```diff
@@ -0,0 +1,14 @@
+import React, { Component } from 'react'
+import TodoList from './components/task-list'
```
jalal75: it is batter to name it as the file name, it will be more readable.

week1-HW/todo-app/src/components/task-list.js
```diff
@@ -0,0 +1,16 @@
...
+export default class TodoList extends Component{
+    render() {
+        return (
+            <ul class="list-group">
+               <li class="list-group-item list-group-item-action list-group-item-primary"> <Tasks todo="Attend HYF less" taskTime="2019-01-13" /></li>
+               <li class="list-group-item list-group-item-action list-group-item-secondary"><Tasks todo="get mine pass card from the gemeenta" taskTime="2019-01-14" />   </li>           
+               <li  class="list-group-item list-group-item-action list-group-item-success">  <Tasks todo="what ever" taskTime="2019-01-015" />    </li>          
+               <li class="list-group-item list-group-item-action list-group-item-danger"> <Tasks todo="what ever" taskTime="2019-07-01" />    </li>          
+                
+            </ul>
```
jalal75: the component <Tasks > has got the <li> tag, so why  did you wrap it in another li here?
you have to use className instead of class
it is batter to but all the props data in one array and then loop over it in the render return

PR #307, opened 2019-01-10
Week1

RimSebai:


homework-submission/my-project/src/TodoItem.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+const TodoItem = props =>
+  <div className="todo-item">
+    <p1>{props.description}</p1>
+    <p2>{props.deadlineDate}</p2>
```
jalal75: There is nothing called p1 or p2, i believe you want to create a paragraph then you have to use p tag.

RimSebai:
Thanks for your feedback!
I did correct the code , can you please explain to me why would I create more than two component ?

jalal75:
you have to make more than 2 components because it is required in your homework.
I believe that the main reason for that is to have the facility to understand how can we divided the page to multicomponent and can be reusable anywhere.
so the best way in your homework is to make three new components (for example) for the header, the description and the deadlineDate
 in this case you have three components that you can use it anytime anywhere

homework-submission/my-project/src/App.js
```diff
@@ -0,0 +1,17 @@
+import React, { Component } from 'react';
+import './App.css';
+import Post from './Post';
+import Data from './Data';
+
+class App extends Component {
+  render() {
+    const newData = Data.map(item => <Post key={item.id} description={item.description} deadlineDate={item.deadlineDate} />)
```
jalal75: it is a good improve in your code here, but we don't need to change our data just for this idea because the map method have got an internal key that we can use it so you can save your time by doing the map method  as I mentioned before like this:
Data.map( ( item ,  key ) => < TodoItem description={item.description} deadlineDate={item.deadlineDate}  key={key} />)

RimSebai: II appreciate your suggestion, but I tried to find another solution and I came out with adding id. I could also use any other unique key, right ? like that:
const newData = Data.map(item => <Post key={item.description} description={item.description} deadlineDate={item.deadlineDate} />)

jalal75: what is the reason that you like it.
you change your data just for your functions but your functions have got this unique key so just use it.
in this case your card has a low cost.

PR #308, opened 2019-01-11
Week1

RimSebai:
improved!


homework-submission/my-project/src/Header.js
```diff
@@ -0,0 +1,13 @@
+import React, { Component } from 'react';
+
+class Header extends Component {
+  render() {
+    return (
+      <div>
+        <header>Todo List</header>
+      </div>
```
jalal75: it is work fine here but you don't need a < div > container because you are returning just one element you need a container just if you return more than one element

homework-submission/my-project/src/Description.js
```diff
@@ -0,0 +1,13 @@
+import React, { Component } from 'react';
+
+class Description extends Component {
+  render() {
+    return (
+      <div>
+        <p>{this.props.description}</p>
+      </div>
```
jalal75: it is work fine here but you don't need a < div > container because you are returning just one element you need a container just if you return more than one element

RimSebai: all done!

jalal75: Well done Rim nice work

RimSebai: Thanks!

PR #309, opened 2019-01-11
React HW1

heboshka:


homework-submission/week1/src/Title.js
```diff
@@ -0,0 +1,11 @@
+import React, { Component } from 'react';
+class Title extends Component {
+  render() {
+    return (
+      <React.Fragment>
+        <h1>Todo List</h1>
+      </React.Fragment>
+    );
```
jalal75: I am happy that you try to use React.Fragment, could you explain why did you use it here? what the reason for it? 
I believe it is not needed, but it works

homework-submission/week1/src/App.js
```diff
@@ -0,0 +1,38 @@
...
+    deadline: 'Fri Sep 15 2017'
+  }
+]
+class App extends Component {
+  render() {
+    const todos = todo.map((item) => (<Todo key={item.key} description={item.description} deadline={item.deadline} />));
+    return (
+      <div className="App">
+        <header className="App-header">
+        </header>
```
jalal75: what this components for.....?
you didn't import it and it is empty

jalal75:
well done
thank you for your homework I like your work
But the best way in your homework is to make three new components (for example) for the header and the post
and inside the post you can make one component for the description and one for the deadline
in this case you have 4 components that you can use it anytime anywhere

homework-submission/week1/src/App.js
```diff
@@ -0,0 +1,38 @@
...
+  },
+  {
+    key: 3,
+    description: 'Eat breakfast',
+    deadline: 'Fri Sep 15 2017'
+  }
+]
+class App extends Component {
+  render() {
+    const todos = todo.map((item) => (<Todo key={item.key} description={item.description} deadline={item.deadline} />));
```
jalal75: you can use the internal key of the Map method instead of the item.id.
like so:
todo.map( ( item , key ) => ( < Todo key = { key } description = { item.description } deadline ={ item.deadline }  / > ) );

PR #311, opened 2019-01-13
Homework Week2

kawaraa:


homework-submission/react-week2/src/App.js
```diff
@@ -0,0 +1,50 @@
...
+          "deadline": "2017-09-09",
+          "done": false
+        }
+      ]
+    };
+  }
+
+  markAs(item) {
+    this.state.todos.forEach(el => {
+      if (el.id === item.id) el.done = el.done ? false : true;
```
jalal75: hard to read
don't mix between ES5 and ES6

homework-submission/react-week2/src/App.js
```diff
@@ -0,0 +1,50 @@
...
+    });
+    this.setState({ todos: this.state.todos });
+  }
+
```
jalal75: The loop here is costly, because you are looping over all the data to find one item which you want to change.
try to fine another solution without loop

homework-submission/react-week2/src/App.js
```diff
@@ -0,0 +1,50 @@
...
+  render() {
+    let list = this.state.todos.map((el) => <Todo key={el.id} todo={el} markAs={this.markAs.bind(this)} />)
```
jalal75: why did you ( bind ) here, it is batter to do it in the constructor

PR #312, opened 2019-01-14
Week2

RimSebai:


homework-submission/my-project/src/App.js
```diff
@@ -0,0 +1,44 @@
...
+          todo.done = !todo.done;
+        }
+        return todo
+      })
+      return {
+        todoData: updatedData
+      }
+    })
+  }
+
```
jalal75: The loop here is costly, because you are looping over all the data to find one item which you want to change.
and you have in line 33 a key in the Map loop, so you can pass this key as a prop to the < Post > component and use it to find the item

homework-submission/my-project/src/App.js
```diff
@@ -0,0 +1,44 @@
...
+  render() {
+    const text = this.state.todoData.length === 0 ? "No item..." : null
+    const newData = this.state.todoData.map((item, key) => <Post key={key} item={item} changeHandel={this.changeHandel} />)
+    return (
+      <div className="App" >
+        <Header/>
+        {text}
+        {newData}
+      </div>
```
jalal75: It is stronger to do it like :
{ this.state.todoData.length === 0 ? "No item..." : newData }

jalal75:
nice

homework-submission/my-project/src/App.js
RimSebai: how could I use newData before it was defined ?

homework-submission/my-project/src/App.js
RimSebai: Sorry, I did't get it

jalal75: you don't need to make a loop because you have the index of the item in your array that mean you can reach the item directly

homework-submission/my-project/src/App.js
jalal75: it was defined in line 33, so you can use it.
just replace the code in line 37 and 38 with my code and it will work fine.

RimSebai: done!

homework-submission/my-project/src/App.js
RimSebai: Sorry for being late. Done

RimSebai:
Thanks for your feedback!

PR #314, opened 2019-01-16
week2

karamad10:


homework-submission/week2/my-app/src/components/ToDoItems.js
```diff
@@ -0,0 +1,51 @@
...
+
+   render() {
+      const blank = <h2>No items...</h2>;
+      return (
+         this.state.Todos.length === 0 ? blank :
+            <div>
+               {
+                  ToDos.map((item, id) => (
+                     <Post
+                        checkbox={<input type="checkbox" checked={item.completed} onChange={() => this.onChange(item.id)} />}
```
jalal75: you are passing the HTML element as prop and that is not good.
you have you create this element in the Post component.

homework-submission/week2/my-app/src/components/ToDoItems.js
```diff
@@ -0,0 +1,51 @@
...
+         const updatedTodos = prevState.Todos.map(todo => {
+            if (todo.id === id) {
+               todo.completed = !todo.completed
+            }
+            return todo
+         })
+         return {
+            Todos: updatedTodos
+         }
+      })
```
jalal75: The loop here is costly, because you are looping over all the data to find one item which you want to change.
and you have in line 35 a key in the Map loop, so you can pass this key as a prop to the < Post > component and use it to find the item

homework-submission/week2/my-app/src/components/ToDoItems.js
```diff
@@ -0,0 +1,51 @@
...
+import Post from "../Post"
+
+import ToDos from "../Todos"
+
+class TodoItem extends Component {
+   constructor(props) {
+      super(props)
+      this.state = {
+         Todos: ToDos
+      }
```
jalal75: why do you put your data ( Todos ) here, it should be in the App file and pass whatever you need to here as props

homework-submission/week2/my-app/src/Post.js
```diff
@@ -0,0 +1,17 @@
+import React, { Component } from 'react';
+
+
+class Post extends Component {
+
+   render() {
+      return (
+         <div className="todo-item">
+            <div>{this.props.checkbox}</div>
```
jalal75: here you need to great the checkbox as HTML element

PR #315, opened 2019-01-16
Week2

joudi6:


homework-submission/todo-app/src/App.js
```diff
@@ -0,0 +1,48 @@
...
+        if (todo.id === id) {
+          todo.done = !todo.done;
+        }
+        return todo;
+      });
+      return {
+        todos: newState
+      };
+    });
+  }
```
jalal75: The loop here is costly, because you are looping over all the data to find one item which you want to change.
and you have in line 31 a key in the Map loop, so you can pass this key as a prop to the < Post > component and use it to find the item

homework-submission/todo-app/src/App.js
```diff
@@ -0,0 +1,48 @@
...
+  render() {
+    const Data = this.state.todos.map((todo, key) => (
+      <Item todo={todo} key={key} handleChange={this.handleChange} />
+    ));
+
+    return (
+      <div className="App">
+        <Header />
+        {this.state.todos.length === 0 ? (
+          <ul className="noItems">No items...</ul>
```
jalal75: batter to use < p > or < h2 > instead of < ul >

PR #316, opened 2019-01-18
Week2

heboshka:


heboshka:
thanks a lot @jalal75 for your comments i think i fixed all you mentioned

jalal75:
Thanks for your quick response 
Well done Hiba

PR #318, opened 2019-01-24
Homework week3

RimSebai:


homework-submission/my-project/src/App.js
```diff
@@ -0,0 +1,68 @@
...
+import './App.css';
+import data from './data';
+import Header from './Header';
+import Form from './Form';
+import Todo from './Todo';
+import Edit from './Edit';
+
+class App extends Component {
+  state = {
+    editing: false,
```
jalal75: it is a good idea to great this variable to change the mode to edit or not, BUT:
it is not good to create a variable and  then change the type of it even if the code is working
( it is boolean and then changed to number)

homework-submission/my-project/src/Form.js
```diff
@@ -0,0 +1,33 @@
...
+          <label>Deadline </label>
+          <input type="date" name="deadline" required />
+          <input type="submit" value="add" />
+        </form>
+      </div>
+    );
+  }
+}
+
+export default Form;
```
jalal75: I believe that if the name of the file is Addform is better

homework-submission/my-project/src/App.js
```diff
@@ -0,0 +1,68 @@
+import React, { Component } from 'react';
+import './App.css';
+import data from './data';
```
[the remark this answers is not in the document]
RimSebai: data file is not component so why do I have to capitalize first letter ?

homework-submission/my-project/src/App.js
RimSebai: you are right, can I use null instead of false ?

homework-submission/my-project/src/Form.js
RimSebai: improved...thanks for your feedback!

RimSebai:
> The application does not allow to edit more than one todo item in the same time, can you change it.
can you please give me a hint...I tried it different ways but I could't solve it

RimSebai:
Thanks Jalal , I improved my code!

jalal75:
I have noticed that  you pushed you chatApp
it is a good idea to make a discussion about it with other students online or in the class 
Keep hard work
Thanks for your interactions

PR #319, opened 2019-01-24
Week3

Damla78:


homework-submission/week3/src/App.js
```diff
@@ -0,0 +1,116 @@
...
+import './App.css';
+import Description from './ToDo/Description';
+import DeadLines from './ToDo/DeadLine';
+import DoneCheckBox from './ToDo/DoneCheckBox';
+import todos from './todoList';
+/** WEEK 3 */
+import NewToDoComponent from './ToDo/NewToDoComponent';
+import ControlsComponent from './ToDo/ControlsComponent';
+
+let oldDescription = '';
```
jalal75: why do you create this variable
just remove it from everywhere in this file, and your code will work fine

homework-submission/week3/src/App.js
```diff
@@ -0,0 +1,116 @@
...
+    tempTodos[x].done = !tempTodos[x].done;
+    this.setState({
+      todos: tempTodos
+    })
+  }
+
+
+  handleSubmit = (newTodo) => {
+    this.setState((state) => ({
+      todos: [...state.todos, newTodo]
```
jalal75: nice code, i like it 👍 💯

homework-submission/week3/src/App.js
```diff
@@ -0,0 +1,116 @@
...
+    oldDescription = '';
+
+    this.setState(
+      {
+        todos: tempToDos,
+        editMode: -1
+      }
+    )
+    this.setState({ editMode: -1 });
+  }
```
jalal75: you don't need the code from line 66 to 75 
because in CANCEL you don't need to change anything except the editMode, so the code in line 76 is enough.

PR #342, opened 2019-02-22
Yaacob - week1 

SignosRk:


jalal75:
Thanks for you hard work
BUT:
- you use a different data: in your data there is no Done and Id which are required and when ToDoItem is done you should give it a red color 
- you should use CDN React to do your homework NOT by create a React App

SignosRk:
Ok I understand, but we have been given green light from our teacher to use create-react-app. Altho I was not using it and only the CDN, got some bugs on it, I asked help to debug because this all is so new to me and I was told to switch to CRA. Thanks for your review but I would like then to people be more active in our group to avoid this kind of miss aproaches, I let my mentor take a look at my code and he said it was fine. I would like this sunday if you have time to explain me better how to correct my app then, thanks for the review man. @jalal75 oh... and as You saw, I also have a very simpel homework in the folder static_list, I used the CDN there... Well, see you then on Sunday!

PR #346, opened 2019-02-23
wk1-react-c18

mohamad-monir:


homework-submission/wk1-app/src/App.js
```diff
@@ -0,0 +1,52 @@
...
+    </ul>
+  );
+};
+
+const Item = Data.map(elem => {
+  return (
+    <li
+      key={elem.id}
+      style={elem.done === true ? { textDecoration: 'line-through' } : { textDecoration: '' }}
+    >
```
jalal75: it is not good to style in 
it is not recommended to make a direct style, better to create a class in CSS file and use it here.
for example:
make a class redColor in your CSS file and to use it in your JS file like:
< li className={`item ${ done && 'redColor' }` }  .. .. ..>

mohamad-monir: thank you @jalal75 for feedback

PR #350, opened 2019-03-02
Week2

ranjbarreza:


Meno-101:
Hello @ranjbarreza,
You have as I saw a great curiosity of using React hooks 😃

But your it is not working, so I think it needs a bit of work in order of making it working.
What about creating react application and replacing the `src/` from the new one with what you have here?

ranjbarreza:
Hi Jawhar, I'll do what you suggested, but that's weird, it works on my machine!

PR #351, opened 2019-03-03
HW week2

soumaiamamo:


homework-submission/my app/my-app/src/add.js
```diff
@@ -0,0 +1,36 @@
...
+    })
+  }
+  render() {
+    return (
+      <div>
+        <form onSubmit={this.submit}>
+          <span>Add Todo</span>
+          <div>
+            <span>Todo description :</span>
+            <input type='text' id='description' onChange={this.todo}>
```
Meno-101: What about `name` instead of `id` attribute?

names usually are for forms, `id` has another uses as a Css selector and many other things, so what if you need to have another id with the same id name afterwards?

homework-submission/my app/my-app/src/add.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+
+class AddTodo extends React.Component {
+  state = {
+    description: '',
+    done: false
+  }
+  todo = (element) => {
+    this.setState({ [element.target.id]: element.target.value })
```
Meno-101: As the previous comment, this will be `element.target.name` which makes more sense for others.

And you will get more flexibility in your code.

homework-submission/my app/my-app/src/add.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+  render() {
+    return (
+      <div>
+        <form onSubmit={this.submit}>
+          <span>Add Todo</span>
+          <div>
+            <span>Todo description :</span>
+            <input type='text' id='description' onChange={this.todo}>
+            </input>
```
Meno-101: you may have a self close tag here instead of a complete element
```HTML
<h1></h1> <!-- this is an element, it can hold content -->
<img /> <!-- this is a tag, it can't holds any content inside of it. -->
```

homework-submission/my app/my-app/src/TodoList.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+const Todos = ({ todos, deleteTodo, addTodo }) => {
```
Meno-101: `addTodo` is a not defined and not used property, in other case you may expect to have a value and it may have a place confusion.

PR #352, opened 2019-03-03
React HW Week2 -Zerisenay

ZerSolo:


homework-submission/my-app/src/App.js
```diff
@@ -0,0 +1,82 @@
...
+      items: [
+        {
+          name: 'REACT Lesson',
+          deadline: 23 - 3 - 2019,
+          key: Date.now(),
+        },
+        {
+          name: 'Wach movie',
+          deadline: 23 - 3 - 2019,
+          key: Date.now() + 1,
```
Meno-101: `Date.now()` giving a number of how many milliseconds started internet _(my guess or as someones said it's since God created the earth 😅)_.

it happen to be great at that number it giving, and as a unique number to use as an id it's good but I'm sure you can do better generating a unique numbers or strings. you can ask how for a short answer or you may google it for more basic knowledge :)

*Either ways feel free to ask*

homework-submission/my-app/src/App.js
```diff
@@ -0,0 +1,82 @@
...
+    };
+    this.addItem = this.addItem.bind(this);
+    this.deleteItem = this.deleteItem.bind(this);
+  }
+
+  addItem(e) {
+    if (this._inputElement.value !== '') {
+      let newItem = {
+        name: this._inputElement.value,
+        key: Date.now(),
```
Meno-101: Oops, you forgot the deadline 🙄

homework-submission/my-app/src/App.js
```diff
@@ -0,0 +1,82 @@
...
+      };
+      this.setState(PrevState => {
+        return {
+          items: PrevState.items.concat(newItem),
```
Meno-101: this is one of the correct ways of using state 🙌🎉

The most important part is just to keep `setState` deals with it.

homework-submission/my-app/src/App.js
```diff
@@ -0,0 +1,82 @@
...
+        };
+      });
+    }
+    this._inputElement.value = ' ';
+    e.preventDefault();
```
Meno-101: for a coding style purpose only, moving this line to the beginning of the function tells anyone reading this code:
> this is a form's submit element.

homework-submission/my-app/src/App.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  deleteItem(key) {
+    const filteredItems = this.state.items.filter(item => {
+      return item.key !== key;
```
Meno-101: you may see some of the confusion here.
people used to use the `id` as the unique identity for any array contains number of objects, and in this case any other developer working on your code won't guess he have to use the `key` as a uniquely identifying property

So in a bigger program _(which you will do)_ this may distract you from other things as well _(you may use key for another purpose)_ and you will have an `id` from the server.

homework-submission/my-app/src/App.js
```diff
@@ -0,0 +1,82 @@
...
+        <div className="header">
+          <h1> Todo APP </h1>
+          <form onSubmit={this.addItem}>
+            <label>
+              TodoItem:
+              <input
+                type="text"
+                name="name"
+                value={this.state.name}
+                ref={a => (this._inputElement = a)}
```
Meno-101: Nice use of `refs` But for a security reason's you may think to use a [controlled input](https://reactjs.org/docs/forms.html) instead, but your work is correct. 💯

homework-submission/my-app/src/Components/button.js
```diff
@@ -0,0 +1,18 @@
+import React, { Component } from 'react';
+
+class Button extends Component {
+  render() {
+    return (
+      <button
+        onClick={this.props.onClick}
+        type={this.props.type}
+        className={this.props.className}
+        key={this.props.key}
```
Meno-101: you may have another look here, the reason: 
 - you have specified a key property to pass it to this component, inside the component this button doesn't receive any property called `key` because it's a stricted prop-name for the react framework for using it in an arrays.

homework-submission/my-app/src/App.js
```diff
@@ -0,0 +1,82 @@
...
+          key: Date.now() + 2,
+        },
+      ],
+    };
+    this.addItem = this.addItem.bind(this);
+    this.deleteItem = this.deleteItem.bind(this);
+  }
+
+  addItem(e) {
+    if (this._inputElement.value !== '') {
```
Meno-101: This is one of the basic use of an if statement.
But now we are an advanced people a bit, So for legibility what about:
```
// because these are applied to both case either a true condition or a false one
this._inputElement.value = ' ';
e.preventDefault();
// the condition goes here...
if (/* this._inputElement.value !== '' // as you have */) {
  // the case of a true condition
}
```

homework-submission/my-app/src/App.js
```diff
@@ -0,0 +1,82 @@
...
+        name: this._inputElement.value,
+        key: Date.now(),
+      };
+      this.setState(PrevState => {
+        return {
+          items: PrevState.items.concat(newItem),
+        };
+      });
+    }
+    this._inputElement.value = ' ';
```
Meno-101: in the condition you have `''` if empty string apply otherwise move on to the next line, 
here you are telling `this._inputElement.value` is equals to white space.

ZerSolo:
Thank you Jawar for your feedback. I really appriciate it a lot.
Accordingly, I will correct the codes you mentioned. I wil let you know if
i get stacked.

PR #355, opened 2019-03-09
Homework week3

SignosRk:
Like my last commit says, the backend needs work, I just got the essential methods working properly which are GET and GET:ID, the other like POST and DELETE, they work but are not directly working with the client, still need to read more documentation to learn how to make the connection between the methods and my components but I am lacking a lot of time. . Would like to learn how for this sunday. Thanks in advance.


homework-submission/react-hyf/react-hyf-client/src/App.js
```diff
@@ -0,0 +1,67 @@
...
+
+  fetchTodos = async () => {
+    const response = await fetch('/todos');
+    const data = await response.json();
+    return data;
+  };
+
+  addTodo(todoText) {
+    let todos = [...this.state.todos];
+    todos.push({ id: this.state.nextId, text: todoText });
```
Meno-101: with this line you are mutating the state directly, you may use instead `.concat` or including it in the previous line as:
```javascript
// Option #1
let todos = [...this.state.todos, { id: this.state.nextId, text: todoText }];

// Option #2
let todos = this.state.todos.concat({ id: this.state.nextId, text: todoText });
```

- You may tell: But this is a shadow copy!

The answer is Well yes it's But it will work only if you have a one level array as `['hello', 567, 'this will be copied']` because this is a reference and you are copying the first layer.

While `[{name: "jawhar"}]` it won't work because it copied the reference that refers to the Object that includes `name: "jawhar"`.

homework-submission/react-hyf/react-hyf-client/src/App.js
```diff
@@ -0,0 +1,67 @@
...
+    this.setState({
+      todos: todos,
+      nextId: this.state.nextId + 1,
```
Meno-101: As the previous Id comment, if you used an id generator the only place you will deal with it is here.

homework-submission/react-hyf/react-hyf-client/src/components/TodoInput.js
```diff
@@ -0,0 +1,34 @@
...
+    if (todo.length > 0) {
+      this.props.addTodo(todo);
+      this.setState({ value: '' });
+    }
+  }
+
+  render() {
+    return (
+      <div>
+        <input type="text" value={this.state.value} onChange={this.handleChange} />
```
Meno-101: You have here a `div` element containing the elements below it, and you have as a parent a `ul` if we have an error Boundary surround the application this will cause an error and the application may never run.

you can find the valid children to `ul` under the example as a permitted content label:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul

PR #362, opened 2019-04-03
Soolaf ,homework week1

soolaf:
Week1 homework
https://codepen.io/Soolaf/pen/NmxmrM


nick-barth:
Good, but you stored data as a global variable, we don't like global variables.  Next time put it in the state of the `<App />`

PR #374, opened 2019-04-11
Utku / Homework Week 2

u61:
I couldn't have enough time to focus on, that's why I used 'refs' for getting input values instead of using components.
I will try to change it for the sake of react learning but objective of homework has been reached.
I am pushing that early-version for review.

Regards,

### DIFF
--- a/homework-submission/homework-week-2/react.js
+++ b/homework-submission/homework-week-2/react.js
@@ -48,7 +48,7 @@ class App extends React.Component {
         done: false
       };
       todos.push(newTodo);
-      this.setState({ todos: todos });
+      this.setState({ todos });
       e.preventDefault();
     }
   }
--- a/homework-submission/homework-week-2/react.js
+++ b/homework-submission/homework-week-2/react.js
@@ -1,0 +1,125 @@
+/* eslint-disable no-undef */
+/* eslint-disable react/react-in-jsx-scope */
+/* eslint-disable react/jsx-no-undef */
+
+class App extends React.Component {
+  constructor(props) {
+    super(props);
+    this.addTodo = this.addTodo.bind(this);
+  }
+
+  state = {
+    todos: [
+      {
+        id: 1,
+        description: 'Get out of bed',
+        deadline: '2017-09-11',
+        done: true
+      },
+      {
+        id: 2,
+        description: 'Brush teeth',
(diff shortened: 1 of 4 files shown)


homework-submission/homework-week-2/react.js
```diff
@@ -0,0 +1,125 @@
+/* eslint-disable no-undef */
+/* eslint-disable react/react-in-jsx-scope */
+/* eslint-disable react/jsx-no-undef */
+
+class App extends React.Component {
+  constructor(props) {
+    super(props);
+    this.addTodo = this.addTodo.bind(this);
```
nick-barth: we can use arrow functions instead of binding

homework-submission/homework-week-2/react.js
```diff
@@ -0,0 +1,125 @@
...
+      {
+        id: 3,
+        description: 'Eat breakfast',
+        deadline: '2017-09-09',
+        done: false
+      }
+    ]
+  };
+
+  addTodo(e) {
```
nick-barth: `addTodo = e =>` will remove the need to bind

homework-submission/homework-week-2/react.js
```diff
@@ -0,0 +1,125 @@
...
+        todo.done = !todo.done;
+      }
+      return todo;
+    });
+    this.setState({ todos: newTodos });
+  };
+
+  deleteTodo = index => {
+    const { todos } = this.state;
+    delete todos[index];
```
nick-barth: nooooooo, don't use delete, it mutates the array, use concat, or splice to not mutate the original array.

homework-submission/homework-week-2/react.js
u61: Has been changed.

homework-submission/homework-week-2/react.js
u61: Replaced by splice but, I can't understand the logic. Why it must be non-mutating ? and splice method is a mutating method as well.

u61: Replaced by splice method but, I can't get the logic. Why it must be non-mutating ? Splice method is a mutating one as well.

u61:
Thank you for review Nick.

PR #380, opened 2019-04-16
Utku / Homework Week 3

u61:
Hi Nick,

I have built my react app.
I have pushed 30 files to keep build healthy.
I will make some improvements following days.

Regards


u61:
I added actually but not staged .gitignore on push.
This is for testing purposes.

PR #383, opened 2019-04-19
homework of week2

ufukcalimli:
Finally I managed it to push. 
Problems are I have `ul` list and the items of it but when it renders I only get the bullets. So I couldn't check `done` feature. Sorry for being late. Week3 is coming soon.


paulbremer:
Hi @ufukcalimli! I took a quick look at your code, the main problem is that the input field for the done state is a textbox and you are checking if `todo.done` is true but it will never be true because its a string not a boolean. You are also using children in the `<Item>` component, but you need to use the `props.children` if you want to render the children.

```
const li = (props, crossed, handleDone) => {
  return (
    <li className={crossed} onClick={handleDone}>
      {props.children}
    </li>
  );
};
```

Take another look at the done state with these pointers and let me know if you have any more questions! 💪🏼

ufukcalimli:
`ul` list is almost done except click functionality. When click on any list item it throws an error that says "TypeError: todos.map is not a function." I couldn't figure it out.

paulbremer:
@ufukcalimli 

So a couple of changes, i'm gonna try to explain it step by step so you learn the most of the steps i took.

In the `render` of the `todoList` I changed the onClick to the following: `onClick={() => this.toggleDone(index)}`. You are already mapping over the array so you have a `index`, better to add it to the toggleDone function than setting it again inside the function.

And here is the new `toggleDone`:
```
toggleDone = index => {
    const { todos } = this.state;

    const newTodos = todos.map((todo, i) => {
      if (index === i) {
        todo.done = !todo.done;
      }
      return todo;
    });
    this.setState({ todos: [...newTodos] });
  };
```

A few weird things you were trying, `todos` is an array inside `this.state`, no need to spread the state as on object (`{ ...this.state }`). Because we are giving the `index` with the function we don't need the `value` and `index` of the target anymore. While mapping over the array for the `newTodos` we can set the todo which matches the `index` to the opposite of it's own value (`todo.done = !todo.done`).

And the final part is `setState`, again we are already mapping over the array with `map` which return us an array, no need to spread it while using `setState`. And don't try to make an Object from the array which will break things! (`this.setState({ todos: { ...newTodos } });`)

It works on my machine 😅, try to fix these things and I hope it will work on your machine too!

ufukcalimli:
Finally I've managed to finish all tasks, I think. Site doesn't look so fancy but functional at least. Thanks for all your helps @paulbremer

paulbremer:
Yes looking good @ufukcalimli! 👍

PR #384, opened 2019-04-21
Week3-Homework - Bassam

BassamHajar:
regarding the request (1): I'm not sure if you want me to add the last homeworks as I did by adding the old components in a folder named (old), and commented out the embedded <PreviousHomeworks/> in the App.js.

for the request (2): the output requested has been presented similarly by refreshing the page to display the next random user data fetched, not by pressing space bar as the targeted does.


react-hyf/server.js
```diff
@@ -0,0 +1,21 @@
...
+const path = require('path');
+const BUILD = './client/build';
+
+const PORT = 8000;
+
+const app = express();
+app.use(express.static(path.join(__dirname, BUILD)));
+
+app.get('/', (req, res) => {
+  res.json("I'm working...");
```
paulbremer: So the root '/' gives a text response (I'm working...) but should give the index.html as a response. Did you get it working? 

Take another look at the task in the homework; (3) a GET route that sends the index.html file from your public folder in the client as a response to the browser (take the / route).

react-hyf/client/src/components/week3_homework/user/User.jsx
```diff
@@ -0,0 +1,36 @@
...
+        }
+      </div>
+    );
+  }
+}
+
+// STYLES
+// user_info_list_style
+const user_info_list_style = {
+  width: '15%',
```
paulbremer: 15% is pretty small on my big screen, this could be way bigger!

PR #385, opened 2019-04-24
Yasser => Week-3  Homework

YaseirAlkhwalda:


homework-submission/react-hyf/react-hyf-client/src/App.css
```diff
@@ -0,0 +1,121 @@
+body {
+  background-image: linear-gradient(rgba(255, 255, 255, 0.4), rgba(255, 255, 255, 1)),
+    url('https://i0.wp.com/www.thenextfact.com/wp-content/uploads/2017/10/reactjs.png?fit=1920%2C1080');
+  background-attachment: fixed;
+  background-position: center;
+  background-repeat: no-repeat;
+  background-size: cover;
+  position: absolute;
```
paulbremer: `position: absolute;` on the `body` is bad practice, remove the absolute from the body, and remove the `left: 40%` from the `.App` and your site will already be more responsive!

homework-submission/react-hyf/react-hyf-client/src/components/HomeworkWeek3/HomeworkWeek3.js
```diff
@@ -0,0 +1,50 @@
...
+const url = 'https://uinames.com/api/?amount=10';
+
+class HomeworkWeek3 extends Component {
+  constructor(props) {
+    super(props);
+    this.state = { data: [] };
+  }
+
+  componentDidMount() {
+    fetch(url)
```
paulbremer: What if the request fails? Make sure you add a `catch` for the errors.

homework-submission/react-hyf/react-hyf-client/src/components/HomeworkWeek3/HomeworkWeek3.js
```diff
@@ -0,0 +1,50 @@
...
+      .then(response => response.json())
+      .then(json => 
+        json.map((elem)=>
```
paulbremer: You are setting the state inside the `map` function which will trigger the function 10 times and only set the data in state to 1 item. It's better to use `setState` once, for example; `this.setState({ data: json[0] })`, since the array is already random it does not matter that we take the first element from the array.

homework-submission/react-hyf/react-hyf-client/src/components/HomeworkWeek3/HomeworkWeek3.js
```diff
@@ -0,0 +1,50 @@
...
+          this.setState({
+            data: elem,
+          })
+        ) 
+    );
+  }
+
+  componentWillMount() {
```
paulbremer: `componentWillMount` is deprecated, use `componentDidMount` instead! https://reactjs.org/docs/react-component.html#unsafe_componentwillmount

homework-submission/react-hyf/react-hyf-client/src/components/HomeworkWeek3/HomeworkWeek3.js
```diff
@@ -0,0 +1,50 @@
...
+    document.addEventListener("keydown", this.onKeyPressed.bind(this));
+  }
+
+  onKeyPressed(e) {
+      if (e.keyCode === 32) {
+        this.componentDidMount();
```
paulbremer: Nice! But you should extract the fetch to it's own function and call that function instead of calling `componentDidMount`

homework-submission/react-hyf/react-hyf-client/src/components/HomeworkWeek3/HomeworkWeek3.js
```diff
@@ -0,0 +1,50 @@
...
+    };
+  }
+
+  render() {
+    const { data } = this.state;
+   
+    return (
+      <div>
+        <Profile
```
paulbremer: Nice abstraction of components!

homework-submission/react-hyf/react-hyf-client/src/components/HomeworkWeek3/Profile.js
```diff
@@ -0,0 +1,26 @@
...
+import Gender from './Gender';
+import Region from './Region';
+import FullName from './FullName';
+
+const Profile = ({ gender, region, name, surname }) => {
+  return (
+    <div>
+      <table className={'profile-table'}>
+        <tbody>
+          <tr className={'tr'}>
```
paulbremer: `tr` as a classname on `tr` is not good practice, give it a better name; `profile-row-fullname`

PR #396, opened 2019-05-01
homework week3

soolaf:


homework-submission/react-hyf/react-hyf-client/src/App.css
```diff
@@ -0,0 +1,99 @@
...
+
+  font-size: 20px;
+
+  text-align: center;
+
+  margin-top: 30px;
+}
+.list {
+  text-align: center;
+  border-radius: 20%;
```
paulbremer: Adding a `border-radius` but no border, that doesn't do anything, either remove it or add a `border`!

homework-submission/react-hyf/react-hyf-client/src/components/Homeworkweek3/Gender.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+const Gender = ({ gender }) => {
+  return (
+    <td>
```
paulbremer: `<td>` needs to be inside a `table` but you are using it inside a listitem (`li`). This also gives you a warning in the dev tools `<td> cannot appear as a child of <li>`. Check your dev tools before pushing code.

homework-submission/react-hyf/react-hyf-client/src/components/Homeworkweek3/index.js
```diff
@@ -0,0 +1,54 @@
...
+    super(props);
+    this.state = {
+      error: null,
+      isLoaded: false,
+      items: [],
+    };
+  }
+
+  componentDidMount() {
+    fetch('https://uinames.com/api/?amount=10')
```
paulbremer: What if the request fails? Make sure you add a `catch` for the errors.

homework-submission/react-hyf/react-hyf-client/src/components/Homeworkweek3/index.js
```diff
@@ -0,0 +1,54 @@
...
+
+  render() {
+    if (!this.state.isLoaded) {
+      return <div>Loading...</div>;
+    } else {
+      return (
+        <div>
+          <ul>
+            {this.state.items.map((item, i) => (
+              <div>
```
paulbremer: There is no need for these `<div>`'s, to only clutter your HTML. It is also giving you an error (`Each child in a list should have a unique "key" prop.`) because the key is on the list and the `<div>` is the first element and the first element should have a key.

soolaf: @paulbremer Thanks for your feedback but the server file still not working

paulbremer: Did you run `npm run build` in the client to build the index.html and then run the server?

soolaf: ja, It works a lot of thanks to you

PR #400, opened 2019-05-08
React HW Week4 - fatih

kfth:
To install dependencies run command: yarn or yarn install
To run the app run command: yarn start


SarKurd:
The works great but the script need to be improved.

if it was my app i'd separate the development and production script like this may be 

```
"client": "cd react-hyf-client && yarn start",
"server": "cd server && nodemon server.js",
"dev": "concurrently --kill-others-on-fail \"yarn server\" \"yarn client\"",
"start": "cd server && node server.js"
```

homework-submission/react-hyf/react-hyf-client/src/App.js
```diff
@@ -0,0 +1,49 @@
...
+import NavBar from './components/NavBar';
+import Week3 from './components/users';
+
+class App extends React.Component {
+  state = {
+    navElements: {
+      Home: { path: '/', description: '' },
+      'Week 1': { path: '/week1', description: 'Static Todolist' },
+      'Week 2': { path: '/week2', description: 'Dynamic Todolist' },
+      'Week 3': { path: '/week3', description: 'Lifecyle methods and fetch api' },
```
SarKurd: I don't think `navElements` should be a state since we never update them.
you could just use `navElements = {...}` outside of state and in NavBar and Home component `navElements={this.navElements}`

homework-submission/react-hyf/react-hyf-client/src/App.js
```diff
@@ -0,0 +1,49 @@
...
+    fetch('https://uinames.com/api/?amount=10')
+      .then(response => response.json())
+      .then(result => this.setState({ data: result, isLoading: false }))
+      .catch(error => this.setState({ error: error.message }));
+  }
+  render() {
+    return (
+      <BrowserRouter>
+        <div>
+          <NavBar navElements={this.state.navElements} />
```
SarKurd: I'd put `BrowserRouer` in `index.js` so we can have access `withRouter` props in `App.js` too

PR #401, opened 2019-05-09
Week4 Homework

AhmadAbdulhaleem:


SarKurd:
You works but there are a few thing to be changed 

1- There is no `react-router-dom` package inside `package.json`, how did you manage to test it without having it inside `package.json` and install it?!!!!

2- why did you push package-lock.json inside root directory?!!!! you were supposed to work inside yor folder only inside `homework-submissions`, please use `git status` before you commit anything so you know what you are pushing

3- `week1` folder inside `homework-submissions` is not required, for week4 you should have removed that folder.

homework-submission/react-hyf/server/server.js
```diff
@@ -0,0 +1,16 @@
+const express = require('express');
+const path = require('path');
+const buildFolder = '../client/build';
+const app = express();
+
+app.use(express.static(path.join(__dirname, buildFolder)));
+
+app.get('/', (req, res) => {
+  res.sendFile(express.static(path.join(__dirname, buildFolder, 'index.html')));
```
SarKurd: it should be `/*` otherwise only root url works.
if you now go to `localhost:4000/users` it doesn't work, server handles it.

homework-submission/react-hyf/server/server.js
```diff
@@ -0,0 +1,16 @@
...
+});
+
+const PORT = 4000;
```
SarKurd: Use `const PORT = process.env.PORT || 4000` otherwise you won't be able to host it.

Host providers return a dynamic port

homework-submission/react-hyf/client/src/routes.js
```diff
@@ -0,0 +1,21 @@
...
+import { Route, Switch } from 'react-router-dom';
+import Week1 from './components/week1-homework/week1';
+import Week3 from './components/week3-homework/week3';
+import Week2 from './components/week2-homework/week2';
+import NotFound from './components/404/not-found';
+import Home from './components/Home';
+
+const Routes = props => {
+  return (
+    <Switch>
```
SarKurd: Routes is a component so you can put it inside components folder.

homework-submission/react-hyf/client/src/routes.js
```diff
@@ -0,0 +1,21 @@
...
+      <Route path="/week1" exact component={Week1} />
+      <Route path="/week2" exact component={Week2} />
+      <Route path="/week3" exact component={Week3} />
```
SarKurd: You need exact only for `/` route or a matching route for example `/posts` and `/posts/:id`
documentation: https://reacttraining.com/react-router/web/api/Switch

homework-submission/react-hyf/client/src/hoc/Layout/layout.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import NavBar from '../../components/Navbar';
+
+const Layout = props => {
+  return (
+    <div>
```
SarKurd: `Layout` isn't a `HOC`, it's a basically a stateless component.
i would make a `layout` folder inside `src` or `components` and put `Layout.js` there

HOC: https://reactjs.org/docs/higher-order-components.html

homework-submission/react-hyf/client/src/components/week3-homework/week3.js
```diff
@@ -0,0 +1,51 @@
...
+  handleKeyPress = e => {
+    if (e.key === ' ' || e.key === 'Spacebar') {
+      fetch(' https://uinames.com/api/?amount=10')
+        .then(res => res.json())
+        .then(result => {
+          const amount = result.slice(1, 2);
+          this.setState({ users: amount, loading: false, show: false });
+        })
+        .catch(err => {
+          console.log(err.message);
```
SarKurd: you have to update the state here in case fetch throws an error, otherwise the user will always see a spinner in case of error too

AhmadAbdulhaleem: all issues have been fixed

PR #403, opened 2019-05-12
week4

Ayh0o0m:


homework-submission/routing/src/components/ApiCall/components/Users.js
```diff
@@ -0,0 +1,46 @@
...
+      <li>
+        {title}: {content}
+      </li>
+    );
+
+    return (
+      <div>
+        <ul>
+          {this.state.people.map((user, i) => (
+            <div>
```
paulbremer: If you map over items in an array you need the set a `key`, but the key should be on the first element inside the map function; in this case the `<div>` element, setting the key on the components inside the div still gives you errors. Try adding the key to the div element instead of the components.

homework-submission/routing/src/components/DynamicToDoList/DynamicToDoList.js
```diff
@@ -0,0 +1,14 @@
+import React, { Component } from 'react';
+import './index.css';
+import TodoList from './TodoList';
+
+class DynamicToDoList extends Component {
+  render() {
+    return (
+      <div>
```
paulbremer: No need for the extra div element, this could be written as `return <TodoList />;`

PR #404, opened 2019-05-12
Week4

ufukcalimli:


homework-submission/week4/week4-project/src/App.js
```diff
@@ -0,0 +1,41 @@
...
+          <Route path="/week1" component={Week1} />} />
+          <Route path="/week2" component={Week2} />
+          <Route path="/week3" component={Week3} />
+        </Switch>
+      </React.Fragment>
+    );
+  }
+}
+
+export default withRouter(App);
```
paulbremer: You don't need to use `withRouter` here, `export default App` is fine. Using `withRouter` gives you access to properties of the Route. You can read more over here https://reacttraining.com/react-router/core/api/withRouter

PR #406, opened 2019-06-24
Enes React HMW1

eneserdek:


homework-submission/index.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <title>TodoApp</title>
+    <script src="https://unpkg.com/react@16/umd/react.development.js"></script>
+    <script src="https://unpkg.com/react-dom@16/umd/react-dom.development.js"></script>
+    <script src="https://unpkg.com/babel-standalone@6.15.0/babel.min.js"></script>
```
Meno-101: It's good that these scripts are a library scripts, other ways they may not work properly.

homework-submission/index.js
```diff
@@ -0,0 +1,75 @@
+/*to create a list item <li> element whose props are description and deadline*/
+
+const ListItem = ({ description, deadline, className }) => {
```
Meno-101: Nice using of stateless component.

homework-submission/index.js
```diff
@@ -0,0 +1,75 @@
...
+    <li className={className}>
+      Description: {description} Deadline: {deadline}
+    </li>
+  );
+};
+
+const StaticList = () => {
+  return (
+    <ul>
+      <p>This is the Static List</p>
```
Meno-101: Paragraph in a `ul` tag may not be a good choice, If I were you I would probably think about that again.

Let's say you have a program with lots of lists and items, and it's an ordered list (or an un-ordered):

 - javascript will look to the `ul` tag to see what to do  with it, and it figures out it suppose to add numbers or pullets to each element as convention of a list
 - then javascript will look to each item on it's own to recognize either it's a list item or it's a sub list
 - if a list item found it will put the number/pullet otherwise it will start another list with other kind of a (pullet/number) style

for anything other than these it won't be able to figure out what is happening with it's list items and sub-lists, and in some programs it **_may_** ignore the whole convention it follows.
Css may act the same as that, especially if you are in a team with lots of people looking to each line and trying to figure out why is this line isn't acting as they asked for?

for more details about how you can know which you can put where, please look the documentation:
 - at the [Permitted content](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul) part below the example given.

homework-submission/index.js
```diff
@@ -0,0 +1,75 @@
...
+      description: 'Eat breakfast',
+      deadline: '2017-09-09',
+      done: false,
+    },
+  ];
+  return (
+    <ul>
+      <p>This is the Dynamic List</p>
+      {dynamicData.map(({ id, description, deadline, done }) => {
+        return (
```
Meno-101: perfect!
but same Permitted content thing here applies on line 45.

homework-submission/index.js
```diff
@@ -0,0 +1,75 @@
...
+    <div>
+      <StaticList />
+      <DynamicList />
+    </div>
+  );
+};
+
+class ContainerList extends React.Component {
+  render() {
+    return <Container_List />;
```
Meno-101: I confused 😕
you have both names `Container_List` and `ContainerList` which makes it a big deal for you in larger applications.

if no need for this Component, you can remove it, but just no need for making similar names.

homework-submission/index.html
eneserdek: For the timebeing, don't know any other way to use react and babel in my project:D

homework-submission/index.js
eneserdek: I deleted unnecessary Container_List variable and write down dynamic and static list directly into the ContainerList. Less code better readibility!

homework-submission/index.js
eneserdek: <p> tag is deleted.

PR #407, opened 2019-06-24
Muhannad - React Homework - week1

Muhannad1989:


homework-submission/homework-week1/index.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8" />
+    <title>Homework week 1 - React</title>
+    <script src="https://unpkg.com/react@16/umd/react.development.js"></script>
+    <script src="https://unpkg.com/react-dom@16/umd/react-dom.development.js"></script>
+
+    <!-- Don't use this in production: -->
+    <script src="https://unpkg.com/babel-standalone@6.15.0/babel.min.js"></script>
```
Meno-101: It's good that these scripts are a library scripts, other ways they may not work properly.
And you have used the same as Sarea's one, good job 👏

homework-submission/homework-week1/app.js
```diff
@@ -0,0 +1,74 @@
+const Item = ({ id, description, deadline, done }) => {
+  // a component has li element and the data which comes from another components(static and dynamic)
+  if (done == true) {
```
Meno-101: A `true/false` are a little confusing for a developer who is trying to reach out the css files and figuring out what belongs to whom.
A css name like would be more convenient as a className by any case.

Another thing you may improve here, is the ternary operator as js1 remember? this is it's place to use!
so you will have:
```javascript
  <li className={done === true ? "crossed-line" : ""}>
    {id} {description} {deadline}
  </li>
```
and that's it no need for any if statement blocks!

Muhannad1989:
Thank you @Meno-101 for your feedback i will try to improve what you mentioned.  thanks

PR #410, opened 2019-06-27
react week1

alaashhadat:


documentation/homework/homeworkweek1/App.js
```diff
@@ -0,0 +1,76 @@
+'use strict';
+
+const ListItem = props => {
+  const { description, deadline } = props.listInfo;
```
Meno-101: Nice using of stateless component.

You have made a re-usable Component that share use for both the dynamic and the static lists, but used it only in the static 🤔

documentation/homework/homeworkweek1/App.js
```diff
@@ -0,0 +1,76 @@
...
+};
+
+const StaticList = props => {
+  return (
+    <div className="static-list">
+      <p>Static List</p>
+
+      <ListItem listInfo={{ description: ' Get out of bed', deadline: 'Wed Sep 13 2017' }} />
+      <ListItem listInfo={{ description: 'Brush teeth', deadline: 'Thu Sep 14 2017' }} />
+      <ListItem listInfo={{ description: ' Eat breakfast', deadline: 'Fri Sep 15 2017' }} />
```
Meno-101: you could use a `ul` instead of the `div` and move the paragraph out side the `ul` making a valid html tree structure. :)

this comment applies here and on the `DynamicList` Component.

documentation/homework/homeworkweek1/App.js
```diff
@@ -0,0 +1,76 @@
...
+};
+
+const DynamicList = props => {
+  const { DynamicListInfo } = props;
+  return (
+    <div className="dynamic-list">
+      <p>Dynamic List</p>
+      {DynamicListInfo.map((item, index) => (
+        <li key={index} className={item.done ? 'done' : 'not-done'}>
+          {item.description},{item.deadline}
```
Meno-101: Two notes are here:
1. It would be possible to re-use `ListItem` instead as the `StaticList` Component since they have the same result: 
```javascript
       <ListItem key={index} listInfo={item} />
```
and it would be nice if you have got the `className` there too 😊

2. It's super clever to solve the problem with the most simplest way, which I like to do, and it's not mandatory for this small application but in a production you will have an `id` for each item just as this one.
the index may cause bugs it's possible to run out of them by using `id` instead.

documentation/homework/homeworkweek1/App.js
```diff
@@ -0,0 +1,76 @@
...
+        id: 3,
+        description: 'Eat breakfast',
+        deadline: '2017-09-09',
+        done: false,
+      },
+    ],
+  };
+
+  render() {
+    console.log(this.state.DynamicList);
```
Meno-101: logging is good but leaving log statements in your code isn't always good for production.

alaashhadat:
Thanks alot for the advice , i will fix it on Monday , have a nice weekend 🙂

Meno-101:
perfect, have a nice one too :)

wouterkleijn:
Thanks for reviewing @Meno-101 . Still awaiting your changes @alaashhadat :)

alaashhadat:
GoodMorning 
Changes can be found in week 2.

PR #411, opened 2019-06-28
Mohammad-Darwesh-homework-react-week1

mohammadnaefaldarwesh:


homework-submission/homeworkWeek1/main.jsx
```diff
@@ -0,0 +1,118 @@
...
+  let { description, deadline, id, done } = listData;
+  return (
+    <li
+      id={id}
+      onClick={() => {
+        if (isStatic) {
+          const element = document.getElementById(id);
+          element.classList.toggle('done');
+        } else {
+          markAsCompleted(id - 1);
```
Meno-101: you have two different functions to define the same job!

For that I would be using another strategy.
Let's say you want to click an element, and hereby we could see the `li` element crossed line.

All we need for that is the element, once we have it we have it's functionality `element.classList.toggle('done');`.
in this Component you have an `id` definition on line 11, using numeric id's, that make it a quite strange for me.
then you are getting it on line 14 with `.getElementById(id)` and it's fine for simplicity sake, but we need simpler.

what about a one function can control both `StaticList` and the `DynaimcList` with no `isStatic` or `markAsCompleted` prop?

it's going to be fun 🙂

**First Step**: We don't need to make a numeric id attribute here!
Please follow the rest of the comments to see what to do next...

homework-submission/homeworkWeek1/main.jsx
```diff
@@ -0,0 +1,118 @@
...
+//This component creates a static list and displays listItem components
+const StaticList = props => {
+  const { isStatic } = props;
+  return (
+    <div>
+      <h2>Static List</h2>
+      <ul className="static list">
+        <ListItem
+          listData={{ id: 1, description: 'Get out of bed', deadline: '13.9.2017' }}
+          isStatic={isStatic}
```
Meno-101: Even if we are going to remove this one `isStatic` but just I liked your way of making things separated and ordered nicely  💯 

Mark This as **Second Step**

homework-submission/homeworkWeek1/main.jsx
```diff
@@ -0,0 +1,118 @@
...
+  return (
+    <div>
+      <h2>Dynamic List</h2>
+      <ul className="dynamic list">
+        {listItems.map(listData => (
+          <ListItem
+            key={listData.id}
+            listData={listData}
+            markAsCompleted={markAsDone}
+            isStatic={isStatic}
```
Meno-101: As I said before we are going to remove `isStatic` and `markAsCompleted` from here.

**Step Three**

homework-submission/homeworkWeek1/main.jsx
```diff
@@ -0,0 +1,118 @@
...
+      },
+    ],
+  };
+
+  markAsDone = index => {
+    let updateState = this.state.dynamicList[index];
+    updateState.done = !updateState.done;
+    this.setState(updateState);
+  };
+
```
Meno-101: make a function here that takes in one parameter, `element` and copy only line 15 of step 1.
this function will be responsible to recognize the element and toggle it's `done` class name.

> it will be helpful to use the `event` Object

And this will be **Step Four**

homework-submission/homeworkWeek1/main.jsx
```diff
@@ -0,0 +1,118 @@
...
+  render() {
+    return (
+      <div className="lists">
+        <StaticList markAsDone={this.markAsDone} isStatic={true} />
+        <DynamicList
+          listItems={this.state.dynamicList}
+          markAsDone={this.markAsDone}
+          isStatic={false}
```
Meno-101: change these two lines 108, 109 and 105 to pass the function you made from **step four**
Make sure to pass it through step 3  and replace with it's all `isStatic` in step 2.

in **step one** change the function you have made there to call this function passing to it the event object from it's caller as 
```javascript
onClick={(event) => { props.myfunction(event) } }
```

homework-submission/homeworkWeek1/main.jsx
```diff
@@ -0,0 +1,118 @@
...
+        deadline: '2017-09-09',
+        done: false,
+      },
+    ],
+  };
+
+  markAsDone = index => {
+    let updateState = this.state.dynamicList[index];
+    updateState.done = !updateState.done;
+    this.setState(updateState);
```
Meno-101: Now no need for this function, But I have a note on this, and it's important.

It's doing the job but it's not the very correct way, this is a very small application and it will not make any kind of affect here, but you will see the affect of that in a future applications.

Line 99 is just changing the state, but more accurately it's been changed before that.
Usually `setState` made for controlling the state in the correct order, so it made it's call and there is a complete lifecycle behind that call, it's not true it's only for rendering!

it have a lot happening behind, in a future lectures of course.
But just let's have our attention on line 98 which is changing the state, if you tried the following it will tell you that this line is changing it:
```javascript
    // if the first log is `true` the other `false` and this behavior tell it's been changed
    console.log(this.state.dynamicList[index].done)
    // this line is changing by reference not by value, which make the 
    updateState.done = !updateState.done;change on the stat by it self
    console.log(this.state.dynamicList[index])
```

perhaps this we will need to try changing the state in a more dynamic way, or keeping it immutable per say, so we could try:
```javascript
const updateState = this.state.dynamicList.map((item) => {
  if (item.id === this.state.dynamicList[index].id) {
    return {
      ...item, done: !item.done
    }
  } else {
    return item
  }
})
```

This is the solution for this thing, I have solved it, because it's a bit tricky and it will be removed either ways.
Please rethink about it, and if you have any question don't keep it, I'm here for help 🙂

homework-submission/homeworkWeek1/main.jsx
```diff
@@ -0,0 +1,118 @@
+'use strict';
+
+const { Component } = React;
```
Meno-101: Happy to see ES6 🎉

Meno-101:
Oh how can I forgot, Very Nice to see it's functionality working though 😃

mohammadnaefaldarwesh:
Hello @Meno-101 @wouterkleijn ,
Thank you very much for your long feedback !....I really appreciate it ....i learned a lot from it but i believe that i did all required task in the homework so i think the `needswork` tag is not necessary  .....please if you think there is something  missing just inform me and i will do it otherwise please remove the `needswork` tag and approve the PR . 
Best regards,

Meno-101:
You're right, It was the best to be honest.
I have used on that button, my bad 😇

PR #412, opened 2019-06-28
Ramazan_React_HW1

RamazanInce:


homework-submission/React_HW1/index.html
```diff
@@ -0,0 +1,155 @@
...
+  <script src="https://unpkg.com/react@16/umd/react.development.js"></script>
+  <script src="https://unpkg.com/react-dom@16/umd/react-dom.development.js"></script>
+
+  <!-- Don't use this in production: -->
+  <script src="https://unpkg.com/babel-standalone@6.15.0/babel.min.js"></script>
+</head>
+
+<body>
+  <div id="root"></div>
+  <script type="text/babel">
```
Meno-101: It would be nice to have an isolated script file.

homework-submission/React_HW1/index.html
```diff
@@ -0,0 +1,155 @@
...
+        {
+          id: 3,
+          description: 'Eat breakfast',
+          deadline: '2017-09-09',
+          done: false,
+        },
+      ];
+
+// TableTitle Component used by both static and dynamic todo list
+      const TableTitle = () => {
```
Meno-101: Great to see a stateless components 🎉

homework-submission/React_HW1/index.html
```diff
@@ -0,0 +1,155 @@
...
+      }
+
+      class StaticTodoContent extends React.Component {
+        render () {
+          return (
+            <tbody>
+              <tr>
+                <td>{todoListStatic[0].description}</td>
+                <td>{todoListStatic[0].deadline}</td>
+                <td>{todoListStatic[0].done}</td>
```
Meno-101: this looks like a static component, I would say it's more to be static content, statics aren't about taking from the parent scope as here, they're about taking from their parents via props.

So this will be a very good example of Components inheritance, 
if you removed this component and tried to use `ContentDynamicTodo` instead in this component's rendering place, it would be as the convention, 

But we could consider the following `Componant` word point's in this case to a class and `content` in my perspective in this case to the `td` element.

And that's why I have called it static content.

homework-submission/React_HW1/index.html
```diff
@@ -0,0 +1,155 @@
...
+      
+// Dynamic todo list used map function component (ContentDynamicTodo) and TableTitle component 
+      const DynamicTodo = () => {
+        return (
+          <div>
+            <h2>Todo list as dynamic</h2>
+           <table>
+              <TableTitle />
+              {todoListDynamic.map(({ id, description, deadline, done }) => ( 
+                <ContentDynamicTodo key={id} description={description} deadline={deadline} className={done ? '√' : 'x'}/>))
```
Meno-101: In line 114 you are using className as `'√'` Or `'x'` which make it a little tricky to catch and follow, I would prefer using English clear words instead, and for line 117 it's looks with styling Great 💯

homework-submission/React_HW1/index.html
RamazanInce: I will do in a separated file. Thanks for that..

homework-submission/React_HW1/index.html
RamazanInce: I did not understand what you mean hear?

homework-submission/React_HW1/index.html
RamazanInce: I try to do i, I hope it is ok.

RamazanInce:
Hi @Meno-101 I add the correction and waiting for your response.

Thanks a lot for everything...

Meno-101:
Great job, And Thanks for your latest commit :)

wouterkleijn:
Super, thanks @RamazanInce and @Meno-101

PR #413, opened 2019-06-29
Wadeea React Homework W1

WadeeaKiwan:


homework-submission/W1-homework/index.js
```diff
@@ -0,0 +1,103 @@
+'use strict';
+
+/**
+ * A function component that inserts a static item in a list
+ * @param {props} param0 The destructured properties description and deadline
+ * @return List item with the values of the properties
```
Meno-101: JSDocs 😎

homework-submission/W1-homework/index.js
```diff
@@ -0,0 +1,103 @@
...
+ * A function component that inserts a dynamic item in a list
+ * @param {props} param0 The array of todoItems
+ * @return List item with the values of the todoItems of the array if todoItem is not done yet (false)
+ * @return List item with the values of the todoItems of the array if todoItem is done (true)
+ * @return The dynamic list of todoItems
+ */
+const DynamicTodoItem = ({ todoItems }) => {
+  const items = todoItems.map(({ id, description, deadline, done }) => {
+    if (!done) {
+      return (
```
Meno-101: Instead of duplicating 4 lines, you could be using `StaticTodoItem` here too, it works!

homework-submission/W1-homework/index.js
```diff
@@ -0,0 +1,103 @@
...
+        <li key={id}>
+          <h3>{description}</h3>
+          <span>{deadline}</span>
+        </li>
+      );
+    }
+    return (
```
Meno-101: you are repeating your self here too, if all of the return content only for the `className` you could consider using ternary operator on line 28 in the `className` instead of this return and the `if` statement as the following:
```javascript
className={done ? "todo-item-done" : ""}
```

And that's it:)

homework-submission/W1-homework/index.js
```diff
@@ -0,0 +1,103 @@
...
+    </li>
+  );
+};
+
+/**
+ * A function component that inserts a dynamic item in a list
+ * @param {props} param0 The array of todoItems
+ * @return List item with the values of the todoItems of the array if todoItem is not done yet (false)
+ * @return List item with the values of the todoItems of the array if todoItem is done (true)
+ * @return The dynamic list of todoItems
```
Meno-101: I will not comment on these beautiful comments _(I like them)_, but Just a one thing to mention.
Here I would use the following as a list instead of repeating `@return` three times:
```javascript
/**
 * A function component that inserts a dynamic item in a list
 * @param {props} param0 The array of todoItems
 * @return
 *  -  List item with the values of the todoItems of the array if todoItem is not done yet (false)
 *  -  List item with the values of the todoItems of the array if todoItem is done (true)
 *  -  The dynamic list of todoItems
 */
```

homework-submission/W1-homework/index.js
WadeeaKiwan: Thanks for this advice ... It has worked also

homework-submission/W1-homework/index.js
WadeeaKiwan: Yes, you are right... I've tried both ways, but this one is better.

homework-submission/W1-homework/index.js
WadeeaKiwan: Thanks... It's done

WadeeaKiwan:
> Well done,
> You have made all required steps correctly,
> I don't have much to say about your homework, but just some comments I have left below.
> 
> And please have a look and ask if you have gotten any questions. 🙂

Many thanks @Meno-101 for your advices and your time.

wouterkleijn:
Super @WadeeaKiwan

PR #416, opened 2019-07-02
homework week1

soolaf:


homework-submission/main.js
```diff
@@ -0,0 +1,67 @@
...
+        <h1>static List</h1>
+        <List description={this.staticList[0].description} deadline={this.staticList[0].deadline} />
+        <List description={this.staticList[1].description} deadline={this.staticList[1].deadline} />
+        <List description={this.staticList[2].description} deadline={this.staticList[2].deadline} />
+
+        <div>
+          <h1>Dynamic List</h1>
+          {this.dynamicList.map(todoList => (
+            <div>
+              <li
```
Meno-101: It's very important to note having an `li` inside of a `div`.

As you can see there is no point of having an `li` inside of a `div`, but the browser even doesn't know how to deal with that situation.

Screen Shot 2019-07-04 at 03 20 26

homework-submission/main.js
```diff
@@ -0,0 +1,67 @@
...
+      done: false,
+    },
+  ];
+
+  render() {
+    return (
+      <div>
+        <h1>static List</h1>
+        <List description={this.staticList[0].description} deadline={this.staticList[0].deadline} />
+        <List description={this.staticList[1].description} deadline={this.staticList[1].deadline} />
```
Meno-101: this comment is same as the previous, the `List` Component is returning an `li` element to a `div` parent.
making it confusing and invalid for the browser.

Please check the `Permitted parents` section of the [mdn docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/li) speaking of the `li` element

homework-submission/main.js
```diff
@@ -0,0 +1,67 @@
...
+        <List description={this.staticList[2].description} deadline={this.staticList[2].deadline} />
+
+        <div>
+          <h1>Dynamic List</h1>
+          {this.dynamicList.map(todoList => (
+            <div>
+              <li
+                key={todoList.id}
```
Meno-101: Having the key anywhere isn't the proper use of it, having it in the very first element of the mapped array _(line 48)_ will remove the following error:

Screen Shot 2019-07-04 at 03 27 30

It's very important to note, Having a key is the React way of identifying each element in the array of it's fellow, And since `.map` method returning an array of elements then it's a valid syntax to say:
```javascript
[
  <p>hello world</p>
]
```

and as a correction for the key thing you may look at the following:
```javascript
// Bad
[
  <div><p key="some unique key">hello world</p></div>
]

// Good
[
  <div key="some unique key"><p>hello world</p></div>
]
```

homework-submission/main.js
```diff
@@ -0,0 +1,67 @@
...
+                style={
+                  todoList.done === true
+                    ? { textDecoration: 'line-through' }
+                    : { textDecoration: '' }
```
Meno-101: Great use of ternary operator here, it's the right place 🎉 

We can improve it as the following:
```javascript
style={ { textDecoration: todoList.done === true ? 'line-through' : '' } }
```

There another improvements we can do, but let's just stick with what you have done 😄

homework-submission/main.js
```diff
@@ -0,0 +1,67 @@
...
+    },
+  ];
+
+  render() {
+    return (
+      <div>
+        <h1>static List</h1>
+        <List description={this.staticList[0].description} deadline={this.staticList[0].deadline} />
+        <List description={this.staticList[1].description} deadline={this.staticList[1].deadline} />
+        <List description={this.staticList[2].description} deadline={this.staticList[2].deadline} />
```
Meno-101: You could have used this component same as here, instead of a new `li` on line 49 with some adjustments, making it dry code.

soolaf: thank you @Meno-101 ,I hope I have corrected it well

homework-submission/main.js
```diff
@@ -38,24 +38,23 @@ class TodoList extends React.Component {
...
-                {todoList.description},{todoList.deadline}
-              </li>
+            <div
+              key={todoList.id}
+              style={
+                todoList.done === true ? { textDecoration: 'line-through' } : { textDecoration: '' }
+              }
+            >
+              {todoList.description},{todoList.deadline}
             </div>
```
Meno-101: Thank you for your changes, no point for any else changes, I will not make any further comments,

but starting from line 51 ending here are exactly the same content of the `List` Component, so you could have used it instead, with the style adjustment on that Component:
```javascript
<List description={todoList.description} deadline={todoList.deadline} />
```

Note: you have to pass the style `object/condition/or even the done property` to the Component via props.

**Good Work!**

PR #417, opened 2019-07-02
Soolaf=> homework week2

soolaf:


homework-submission/main.js
```diff
@@ -0,0 +1,82 @@
+'use strict';
+function AddingForm({ addTodoItem }) {
```
Meno-101: A very Good use of a Form component1

homework-submission/main.js
```diff
@@ -0,0 +1,82 @@
...
+  render() {
+    return (
+      <div>
+        <h1>Dynamic List</h1>
+        <AddingForm addTodoItem={this.addItem} />
+
+        {this.state.dynamicList.map((todoList, index) => (
+          <div className='todoList'>
+            <li
+              key={todoList.id}
```
Meno-101: you have corrected this element in the previous homework, and I have replied with a comment, It would be very nice to see the changes here 😄

And I liked your use of the id this is the right way of identifying keys! 🎉

homework-submission/main.js
```diff
@@ -0,0 +1,82 @@
...
+
+    this.setState({ todoList: newTodoList });
+  };
+  render() {
+    return (
+      <div>
+        <h1>Dynamic List</h1>
+        <AddingForm addTodoItem={this.addItem} />
+
+        {this.state.dynamicList.map((todoList, index) => (
```
Meno-101: This isn't a list, but an item, making your variable names clear will make it easier to read in larger applications.

homework-submission/main.js
```diff
@@ -0,0 +1,82 @@
...
+
+  addItem = element => {
+    element.preventDefault();
+    const newTodo = {
+      id: this.state.dynamicList.length + 1,
+      description: element.target.description.value,
+      deadline: element.target.deadline.value,
+      done: false,
+    };
+    console.log(newTodo);
```
Meno-101: No logs in your current or future projects, It would be a lot cleaner when you remove them after you are done.

homework-submission/main.js
```diff
@@ -0,0 +1,82 @@
...
+          done: false,
+        },
+      ],
+    };
+  }
+
+  addItem = element => {
+    element.preventDefault();
+    const newTodo = {
+      id: this.state.dynamicList.length + 1,
```
Meno-101: This is a solution of making a new key,
perhaps using a random thing would make a better uniquely use of ids, and therefore a unique key for each item in the list, and easier to recognize for react.

this would be a reason for the following message and causing some bugs in the future!
```red
Warning: Encountered two children with the same key, `1`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

you could replace this value with the following 
```javascript
Math.random().toString(36).substr(2, 10)
```
this will make sure it will be as unique as possible, or you could search google to find this way of making a unique string of letters:
 - one of the following solution would fit:
  https://stackoverflow.com/questions/6860853/generate-random-string-for-div-id 
 - or this:
  https://www.w3resource.com/javascript-exercises/javascript-math-exercise-23.php
 - or even installing this library would make it happen 😃
  _it's just an example you don't have to install anything here!_
  https://www.npmjs.com/package/uuid

And I like simplest solutions better.

homework-submission/main.js
```diff
@@ -0,0 +1,82 @@
...
+      description: element.target.description.value,
+      deadline: element.target.deadline.value,
+      done: false,
+    };
+    console.log(newTodo);
+    this.setState({ dynamicList: this.state.dynamicList.concat(newTodo) });
```
Meno-101: Perfect use of `.concat` Method!

homework-submission/main.js
```diff
@@ -0,0 +1,82 @@
...
+  };
+  removeItem = index => {
+    const newTodoList = this.state.dynamicList.splice(index, 1);
```
Meno-101: if you tried the following you will notice a direct change of the state
```javascript
    console.log(this.state.dynamicList.length) // if this 3
    const newTodoList = this.state.dynamicList.splice(index, 1); // you are splicing here!
    console.log(this.state.dynamicList.length) // this is 2 and it should stay 3
```

the `.setState` method isn't only for redering the content, it's made on top of a complete life cycle happening on behalf of your program, and we want to make sure we are not changing the state by hand as here.

the best solution here would be either to make a deep copy of the array, or making a new array, from js2 we remember having `.filter` will return a new array, and this is the best example of it, and where it should go.

homework-submission/main.js
```diff
@@ -0,0 +1,82 @@
...
+    console.log(index);
+
+    this.setState({ todoList: newTodoList });
```
Meno-101: I havn't saw any `todoList` property in your state before this line.

I have three major notes to point here:
 - the new property of `todoList` you are creating and not using at all kind.
 - `newTodoList` variable contains the latest item you have removed only, check it by `console.log(newTodoList)`
 - and the previous comment :)

soolaf: @Meno-101 thank you

Meno-101:
Well done! Thank you for your changes @soolaf.

wouterkleijn:
Well done @soolaf :) Dont forget to ask questions in class ;)

PR #418, opened 2019-07-03
React week2

saidcetinkaya:


homework-submission/week2/App.js
```diff
@@ -0,0 +1,127 @@
...
+      newItem: {
+        id: '',
+        description: 'Description',
+        deadline: 'nn',
+        done: false,
+      },
+    };
+  }
+
+  NewListItemDescription = () => {
```
Meno-101: I see double,
starting from line 58 ending with line 74, is the exact same as starting from line 76 ending with line 92.

The only difference in these two components is one word in line 68 vs line 86 _(description)_ changed to _(deadline)_.

In any kind like of this situation, remember to dry **(Don't repeat yourself)**
to solve this you can do the following:
```javascript
   NewListItem = (props) => {
    return (
      <p>
        <label htmlFor="">Deadline</label>
        <input
          type="datetime-local"
          name={props.name} // since this is a component we can specify the name via props
          onChange={event => {
            const inputValue = event.target.value;

            this.setState({
              newItem: {
                 ...this.state.newItem,
                 // using the name in this kind of use allowing use to modify the key name to be as we specify the value of it
                 [props.name]: inputValue,
                 id: Date.now()
              },
            });
          }}
        />
      </p>
    );
  };
```
Example use of changed key name:
 name is 'description' means our object will look like: {..., description: 'somevalue'}
 same if 'deadline' means our object will look like: {..., deadline: 'some deadline'}

**Don't forgot to pass the name through props**

homework-submission/week2/App.js
```diff
@@ -0,0 +1,127 @@
...
+    return (
+      <p>
+        <label htmlFor="">Description</label>
+        <input
+          type="text"
+          onChange={event => {
```
Meno-101: it's fine to keep a complete function here for _demonstrating_ purposes, but not in a production!
this will be a confusing part for anyone reading the code.

Usually assigning a value will be once and forever, especially when we are dealing with ids, making them very unique to the item and not any other, you are doing that two times first on line 68 and the second on line 86.

and you are doing that on any change so if no changes happens and you wished to have the same exact content of the previous item here is how it goes:

ezgif com-video-to-gif(2)

And note that the console will complain about `Warning: Encountered two children with the same key`

So I would leave this event arrow function here but the id submission in a click event attached to the add button instead!
This will help keeping it:
 1. added once 
 2. submitting as many as wanted on every click since it's a click event 🙂

homework-submission/week2/App.js
```diff
@@ -0,0 +1,127 @@
...
+  };
+
+  addToState = () => {
+    this.setState({
+      toDos: [...this.state.toDos, this.state.newItem],
+    });
+  };
+
+  removeFromState = id => {
+    const newTodos = this.state.toDos.filter(todo => id !== todo.id);
```
Meno-101: perfect using of `.filter` method 💯

wouterkleijn:
Can you address the feedback @saidcetinkaya ?

PR #422, opened 2019-07-05
Enes React HMW2

eneserdek:


homework-submission/App.js
```diff
@@ -0,0 +1,105 @@
...
+    super(props);
+    this.state = {
+      todos: [],
+    };
+  }
+
+  componentDidMount() {
+    readFile(path).then(response => {
+      this.setState(() => ({
+        todos: response,
```
Meno-101: Super!

homework-submission/methods.js
```diff
@@ -0,0 +1,10 @@
+function readFile(path) {
+  return new Promise((resolve, reject) => {
```
Meno-101: You don't need to return a new Promise in order to get the promise!

`fetch` contains the promise you have asked for, I would consider returning it instead!

homework-submission/server.js
```diff
@@ -0,0 +1,14 @@
+const express = require('express');
```
Meno-101: Since you've made an express server based application, Where is your `package.json`?

homework-submission/App.js
```diff
@@ -0,0 +1,105 @@
+const path = '/data.json';
+
+const ListItem = ({ id, description, deadline, className }) => {
```
Meno-101: Less components in react means less readability.

you only have two!

homework-submission/App.js
```diff
@@ -0,0 +1,105 @@
...
+            name="todo_deadline"
+            placeholder="Give a deadline"
+            onChange={event => {
+              NewTodo.deadline = event.target.value;
+            }}
+          />
+          <input
+            type="submit"
+            value="Add"
+            onClick={event => {
```
Meno-101: it would be very complicated in larger apps adding more in line event functions,
perhaps isolating them in a new method could be more legible, shorter and easier to follow.

homework-submission/App.js
```diff
@@ -0,0 +1,105 @@
...
+              TodoNumber = event.target.value;
+            }}
+          />
+          <input
+            type="submit"
+            value="Delete"
+            onClick={event => {
+              event.preventDefault();
+              const index = TodoNumber - 1;
+              const MarkedTodo = this.state.todos[index];
```
Meno-101: When I run the server with a provided `data.json` file this function isn't working.

For simplicity making it without the server for now could help debugging.

homework-submission/App.js
```diff
@@ -0,0 +1,105 @@
...
+            <ListItem
+              id={id}
+              description={description}
+              deadline={deadline}
+              className={done === true ? 'strike_through' : ''}
+              key={i}
+            />
+          ))}
+        </ul>
+        <form>
```
Meno-101: This form can get a Component on it's own

eneserdek:
I will work on comments and try to address all.

homework-submission/methods.js
eneserdek: Relevant changes are made!

homework-submission/App.js
eneserdek: New components are introduced.

homework-submission/App.js
eneserdek: Functions are listed seperately and reference is made while using them.

homework-submission/App.js
eneserdek: Now it is working.

homework-submission/App.js
eneserdek: Divided into components.

homework-submission/App.js
```diff
@@ -1,105 +1,161 @@
...
+    readFile(path)
+      .then(data => data.json())
+      .then(response => {
+        this.setState(() => ({
+          todos: response,
+        }));
+      })
+      .catch(error => {
+        this.setState(() => {
+          todos: error.message;
```
Meno-101: This `.setState` won't work!

homework-submission/App.js
```diff
@@ -1,105 +1,161 @@
...
+        });
+      });
+  }
+  //to add new todos with unique id's. Even though a todo is deleted, none of the id's will be coincided.
+  addTodo(event, NewTodo, todos) {
+    event.preventDefault();
+    const lastTodoIndex = todos.length - 1;
+    NewTodo.id = todos[`${lastTodoIndex}`].id + 1;
```
Meno-101: smart id setting :)

homework-submission/App.js
```diff
@@ -1,105 +1,161 @@
...
+    todos.push(NewTodo);
```
Meno-101: if you tried the following you will notice a direct change of the state
```javascript
    console.log(this.state.todos.length) // if this 3
    todos.push(NewTodo); // you are pushing here!
    console.log(this.state.todos.length) // this is 4 and it should stay 3
```

the `.setState` method isn't only for rendering the content, it's made on top of a complete life cycle happening on behalf of your program, and we want to make sure we are not changing the state by hand as here.

the best solution here would be as the following:
```javascript
    this.setState({
      todos: [...todos, newTodo], // this will be a safe way to add to your state without affecting it.
    });
```

homework-submission/App.js
```diff
@@ -1,105 +1,161 @@
...
+    const updatedList = todos.filter((todo, index) => {
+      return index != TodoNumber - 1;
+    });
+    this.setState({ todos: updatedList });
+  }
   render() {
     const todos = this.state.todos;
+    // this new todo is produced to be pushed when Add new todo button is clicked.
     const NewTodo = {
       id: 0,
```
Meno-101: State's existence for handling this job for us, so why not using it, it will make it a lot easier and familiar for others

wouterkleijn:
Please address the comments @eneserdek

PR #423, opened 2019-07-05
Wadeea react homework week2

WadeeaKiwan:


homework-submission/W2-homework/todo-app/src/Components/AddItem/AddItem.js
```diff
@@ -0,0 +1,51 @@
...
+    this.setState({
+      [e.target.id]: e.target.value,
+    });
+  };
+
+  handleSubmit = e => {
+    e.preventDefault();
+    if (e.target.description.value === '') {
+      return false;
+    } else {
```
Meno-101: If you are returning no need for the `else` close!

the return keyword stops the function in it's exact place and never execute any line after.

homework-submission/W2-homework/todo-app/src/App.js
```diff
@@ -0,0 +1,59 @@
...
+    let todoItems = this.state.todoItems.filter(todoItem => {
+      return todoItem.id !== id;
+    });
+    this.setState({ todoItems });
+  };
+
+  addTodoItem = todoItem => {
+    let todoItems = this.state.todoItems;
+    todoItem.id = Math.random() * 10 + 1;
+    todoItems.push(todoItem);
```
Meno-101: You are manipulating the state directly!

If you tried the following:
```javascript
    console.log(this.state.todoItems.length) // if this 3
    todoItems.push(todoItem)
    console.log(this.state.todoItems.length) // this is changed to be 4, which it should not do that
```

the `.setState` method is on top of a complete lifecycle on behalf of eyes which made react makers make it asynchronous, and that is why the strange behavior you have saw here.

This is an important thing to keep in mind avoiding Bugs in the future, and understanding how react dealing with it's own state.

you could solve this as the following allowing your self to add an item without touching the state directly:
```javascript
this.setState({
  todoItems: [
    ...this.state.todoItems,
    {
      ...todoItem,
      id: /* Some unique id... */
    }
  ]
})
```

homework-submission/W2-homework/todo-app/src/Components/AddItem/AddItem.js
WadeeaKiwan: Thanks... done!

homework-submission/W2-homework/todo-app/src/App.js
WadeeaKiwan: Many thanks for this advice.

WadeeaKiwan:
> Excellent using of `create-react-app`!
> 
> I don't have much to say, I just noticed a one thing can be improved, no other words!
> 
> **Keep it up...**

Many thanks @Meno-101 for your feedback and advices.

PR #424, opened 2019-07-06
Muhannad React week2

Muhannad1989:


homework-submission/homework-week1/app.js
```diff
@@ -0,0 +1,160 @@
+const Item = ({ id, description, deadline, done, removeItem }) => {
+  return (
+    <li key={id} className={done === true ? 'true' : ''}>
```
Meno-101: No need for a key here, while you have a one in it's correct spot at the very first level of the array `.map` generates!
```suggestion
    <li className={done === true ? 'true' : ''}>
```

You have className-s such as `true` or `ok`, they are fine, But I would go for another class names meaningful css-like instead

homework-submission/homework-week1/app.js
```diff
@@ -0,0 +1,160 @@
...
+
+  error() {
+    let element = document.getElementById('error');
+    element.classList.remove('hidde');
+  }
+
+  handelSubmit = e => {
+    e.preventDefault();
+    if (e.target['description'].value !== '' && e.target['deadline'].value !== '') {
+      this.props.addItem(this.state);
```
Meno-101: Excellent use of passed state!

homework-submission/homework-week1/app.js
```diff
@@ -0,0 +1,160 @@
...
+  };
+
+  error() {
+    let element = document.getElementById('error');
+    element.classList.remove('hidde');
+  }
+
+  handelSubmit = e => {
+    e.preventDefault();
+    if (e.target['description'].value !== '' && e.target['deadline'].value !== '') {
```
Meno-101: you are using two ways of handling values:
 - one by state
 - other using the form's target values

It would be shorter and easier to use one of them instead!

homework-submission/homework-week1/app.js
```diff
@@ -0,0 +1,160 @@
...
+        <h2>To Do App</h2>
+        <AddItems addItem={this.addItem} />
+        <h2>To Do List</h2>
+
+        <TodoItems toDoList={this.state.toDoList} removeItem={this.removeItem} />
+        <div className="error hidde" id="error">
+          <div className="message_box ">
+            <p>
+              All fields are required please insert all fields{'  '}
+              <i class="fa fa-exclamation-triangle" />
```
Meno-101: I see a warning in my console, you probably forgot this:
```red
Warning: Invalid DOM property `class`. Did you mean `className`?
```
```suggestion
              <i className="fa fa-exclamation-triangle" />
```

homework-submission/homework-week1/app.js
```diff
@@ -0,0 +1,160 @@
...
+    });
+    this.setState({ toDoList });
+  };
+
+  addItem = item => {
+    const random = Math.random(Math.random() * 9999);
+    item.id = random;
+    let items = this.state.toDoList;
+    items.push(item);
+    this.setState({ items });
```
Meno-101: if you tried the following you will notice a direct change of the state
```javascript
    console.log(this.state.toDoList.length) // if this 3
    items.push(item); // this line is mutating the original state here!
    console.log(this.state.toDoList.length) // this is 4 and it should stay 3
```

the `.setState` method isn't only for rendering the content, it's made on top of a complete life cycle happening on behalf of your program, and we want to make sure we are not changing the state by hand as here.

the best solution here would be as the following:
```javascript
    this.setState({ items: [ ...items, item ] }); // and that how easy it's 😃
```

homework-submission/homework-week1/app.js
```diff
@@ -0,0 +1,160 @@
...
+
+  removeItem = id => {
+    const toDoList = this.state.toDoList.filter(item => {
+      return item.id !== id;
+    });
+    this.setState({ toDoList });
+  };
+
+  addItem = item => {
+    const random = Math.random(Math.random() * 9999);
```
Meno-101: This will solve the problem, but why a random inside a random? 😕

homework-submission/homework-week1/app.js
Muhannad1989: yes,  you are right here. I've forgot to change it .

homework-submission/homework-week1/app.js
Muhannad1989: It was my fault,  I would say random inside round.

Muhannad1989:
hello @Meno-101 , 
Thank you for your feedback , I've fixed all problems you have mentioned.

PR #425, opened 2019-07-06
Week2

alaashhadat:


reacthomeworkweek2/App.js
```diff
@@ -0,0 +1,90 @@
+'use strict';
+
+class Container extends React.Component {
```
Meno-101: you have fit your program in a one Component.
Changing that now will be changing most of your work, so that's fine, I will be expecting more components in the next homework for more legibility, and ise of practice. 🔥

reacthomeworkweek2/App.js
```diff
@@ -0,0 +1,90 @@
...
+        {
+          description: 'Eat breakfast',
+          deadline: '2017-09-09',
+          done: false,
+        },
+      ],
+    };
+  }
+  handleDelete(index) {
+    event.preventDefault();
```
Meno-101: where this `event` came from? and what does it do here? 🤔

reacthomeworkweek2/App.js
```diff
@@ -0,0 +1,90 @@
...
+    const items = [...this.state.items];
```
Meno-101: You have a very good use of a shadow copies 😄

reacthomeworkweek2/App.js
```diff
@@ -0,0 +1,90 @@
...
+
+    items.splice(index, 1);
+    this.setState({ items });
+  }
+  handleSubmit(event) {
+    event.preventDefault();
+    let text = { ...this.state.text };
+    const items = [...this.state.items];
+    text.description && text.deadline ? items.push(text) && this.setState({ items }) : null;
```
Meno-101: An if statement would have a better look here (readability perspective), agree?

reacthomeworkweek2/App.js
```diff
@@ -0,0 +1,90 @@
...
+            name="description"
+          />
+          <label id="title">Deadline:</label>
+          <input id="input" type="date" onChange={this.handleChange.bind(this)} name="deadline" />
+          <button id="add">Add</button>
+        </form>
+
+        {this.state.items.map((item, index) => (
+          <div key={index}>
+            <li className={item.done ? 'done' : 'not-done'}>
```
Meno-101: An `li` inside of a `div` would be great if it works as a list 😅

reacthomeworkweek2/App.js
```diff
@@ -0,0 +1,90 @@
...
+              {item.description},{item.deadline}
+            </li>
+            <li>
+              <a
+                id="delete"
+                href="#"
+                onClick={() => {
+                  this.handleDelete.bind(this);
+                  this.handleDelete(index);
```
Meno-101: you could call these two lines as the following:
```javascript
this.handleDelete.call(this, index) // only this will be the right replaceable for both
```
or using the arrow functions instead of the classical class methods style:
```javascript
  handleDelete = (event) => { // with this way no need for binding any more:)
    //- Code...
  }
```

reacthomeworkweek2/App.js
```diff
@@ -0,0 +1,90 @@
...
+            onChange={this.handleChange.bind(this)}
+            name="description"
+          />
+          <label id="title">Deadline:</label>
+          <input id="input" type="date" onChange={this.handleChange.bind(this)} name="deadline" />
+          <button id="add">Add</button>
+        </form>
+
+        {this.state.items.map((item, index) => (
+          <div key={index}>
```
Meno-101: If you're having an ids using them would a valuable thing
using indexes can cause in bugs.

using indexes is might replaces the ids but only in case of not having them as the [docs](https://reactjs.org/docs/lists-and-keys.html) says:
> When you don’t have stable IDs for rendered items, you may use the item index as a key

And beyond that they're saying:
> We don’t recommend using indexes for keys if the order of items may change. This can negatively impact performance and may cause issues with component state.

alaashhadat:
Hello 
Thank you for the approve. I will keep working on my skill and i will take your advices into account to improve my code.

PR #427, opened 2019-07-07
Orhan_ReactW2_Homework

OrhanGoc:


homework-submission/Week2/App.js
```diff
@@ -0,0 +1,119 @@
...
+      deadline: this.state.newTodo.deadline,
+      done: false,
+    };
+    this.setState({ todoList: [...this.state.todoList, newTodo] });
+  };
+
+  DeleteItem = id => {
+    const { todoList } = this.state;
+    const indexOfItem = todoList.findIndex(elem => elem.id === id);
+    todoList.splice(indexOfItem, 1);
```
Meno-101: if you tried the following you will notice a direct change of the state
```javascript
    console.log(this.state.todoList.length) // if this 3
    todoList.splice(indexOfItem, 1); // you are splicing here!
    console.log(this.state.todoList.length) // this is 2 and it should stay 3
```

the `.setState` method isn't only for redering the content, it's made on top of a complete life cycle happening on behalf of your program, and we want to make sure we are not changing the state by hand as here.

the best solution here would be either to make a deep copy of the array, or making a new array, from js2 we remember having `.filter` will return a new array, and this is the best example of it, and where it should go.

homework-submission/Week2/App.js
```diff
@@ -0,0 +1,119 @@
...
+            />
+          );
+        })}
+      </ul>
+    );
+  };
+
+  RenderForm = () => {
+    return (
+      <form>
```
Meno-101: you are having two kind of elements selection,
 - one by state
 - another using the form's targets names
removing the form will be a good thing, so no need to prevent default behavior in your event any more!

That would make it shorter since you will remove the input names too.

homework-submission/Week2/App.js
```diff
@@ -0,0 +1,119 @@
...
+        <label>New Todo: </label>
+        <input
+          type="text"
+          name="description"
+          value={this.state.newTodo.description}
+          onChange={event => {
+            const inputValue = event.target.value;
+            this.setState({ newTodo: { ...this.state.newTodo, description: inputValue } });
```
Meno-101: it's fine to keep a complete function here for _demonstrating purposes_,but not in a production!
this will be a confusing part for anyone reading the code, especially if it became longer.

homework-submission/Week2/App.js
```diff
@@ -0,0 +1,119 @@
...
+          }}
+        />
+        <input
+          type="date"
+          name="deadline"
+          value={this.state.newTodo.deadline}
+          onChange={event => {
+            const inputValue = event.target.value;
+            this.setState({ newTodo: { ...this.state.newTodo, deadline: inputValue } });
```
Meno-101: Same as previous comment.

homework-submission/Week2/App.js
```diff
@@ -0,0 +1,119 @@
...
+    const { todoList } = this.state;
+    const indexOfItem = todoList.findIndex(elem => elem.id === id);
+    todoList.splice(indexOfItem, 1);
+    this.setState({ todoList: todoList });
+  };
+
+  RenderList = () => {
+    return (
+      <ul>
+        {this.state.todoList.map(props => {
```
Meno-101: calling this in a name other than props will make it much better as a `.map` method for readability.

homework-submission/Week2/App.js
OrhanGoc: I used .filter  as you mentioned.

homework-submission/Week2/App.js
OrhanGoc: Removed the form, but i used names in another function, so names are still there.

homework-submission/Week2/App.js
OrhanGoc: created a new function for both onChange events

OrhanGoc:
Thank you @Meno-101 for your detailed feedback, i learned a lot from you and changed my code as you recommended.

PR #432, opened 2019-07-13
Week3

alaashhadat:


react-hyf/react-hyf-client/src/components/Homeworkweek3.js
```diff
@@ -0,0 +1,55 @@
...
+  componentDidMount() {
+    this.fetchUsers();
+  }
+
+  render() {
+    const { isLoading, users, error } = this.state;
+    return (
+      <React.Fragment>
+        <h1 id="user"> Users</h1>
+        {error ? <p>{error.message}</p> : null}
```
Meno-101: mostly we find this condition where we have to have a `null` value for no reason, or just to say this is a nothing value, it a correct use, but you could do the following too:

```suggestion
        {error && <p>{error.message}</p>}
```
this line now just says:
> if `error` is a truthy value the paragraph will be returned otherwise nothing will be shown, unless a zero since the zero is a falsey value and it used to be visible.

alaashhadat:
Thanks  for the support. Appreciatie it 🙂

PR #433, opened 2019-07-13
Wadeea react homework week3

WadeeaKiwan:


homework-submission/react-hyf/react-hyf-client/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "name": "react-hyf-client",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "react": "^16.8.6",
+    "react-dom": "^16.8.6",
+    "react-scripts": "3.0.1"
+  },
+  "scripts": {
+    "client-install": "cd .. && cd react-hyf-client && npm install",
```
Meno-101: you are going back a directory and going into it again 🤔

homework-submission/react-hyf/server/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "server",
+  "version": "1.0.0",
+  "description": "Express server of react-hyf",
+  "main": "server.js",
+  "scripts": {
+    "start": "node server.js",
+    "server": "nodemon server.js",
+    "client": "cd .. && cd react-hyf-client && npm start",
```
Meno-101: instead of making three sentences we could make it two:
```suggestion
    "client": "cd ../react-hyf-client && npm start",
```

homework-submission/react-hyf/server/server.js
```diff
@@ -0,0 +1,16 @@
+const express = require('express');
+
+const app = express();
+app.use(express.static('../react-hyf-client/build'));
+
+app.get('/', async (request, response) => {
+  await response.sendFile('index.html', err => {
+    throw err;
```
Meno-101: I have to mention two points here:
 1. You are throwing error by any case
 2. `app.use` is doing the exact same thing you are doing in `app.get` :)

homework-submission/react-hyf/react-hyf-client/src/Components/Homework3/Homework3.js
```diff
@@ -0,0 +1,27 @@
...
+import './Homework3.css';
+import { getUsers } from './services/usersServices';
+import UserList from './Components/UserList/UserList';
+
+/**
+ * This class component is the container of the Users List
+ * @return Here is the return of the Users List
+ */
+class Homework3 extends Component {
+  state = {};
```
Meno-101: in a strict mode this will cause an error, which is something you may encounter.
what about initializing the users array here before fetching?

homework-submission/react-hyf/react-hyf-client/package.json
WadeeaKiwan: Yes, you're right... It's my fault, I wanted to put this script in the other package.json of the server folder, so that no need then to navigate to the client folder to install the dependancies. :) But it seems that I've put it in the other one... Corrected!

homework-submission/react-hyf/server/package.json
WadeeaKiwan: Of course this is better ... Thanks for this advice

homework-submission/react-hyf/server/server.js
WadeeaKiwan: Many thanks for these advices ... I've tried them... I still have one question over the first point. How to use this err callback function of `sendFile` in this case?

The `app.get` is deleted as it's doing the same thing of `app.use`

homework-submission/react-hyf/react-hyf-client/src/Components/Homework3/Homework3.js
WadeeaKiwan: Many thanks here also for this advice... Done!

WadeeaKiwan:
> As last homework, Great job!
> 
> I have left for you a few comments, please have a look and tell me your thoughts 🔥

Thanks a lot Jawher @Meno-101 for your time, advices and feedback.... have a nice time :)

PR #437, opened 2019-07-14
RamazanInce_React_HW3_New

RamazanInce:


homework-submission/react-hyf/server/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "start": "node server.js"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "concurrently": "^4.1.1",
+    "express": "^4.17.1",
+    "nodemon": "^1.19.1"
```
Meno-101: you have a `concurrently` and `nodemon` packages with no use, it will take more memory and disk space.

homework-submission/react-hyf/react-hyf-client/src/React_HW2/todoApp2.jsx
```diff
@@ -0,0 +1,183 @@
...
+      </tr>{' '}
+    </tbody>
+  );
+};
+
+// DynamicList where map function is
+const DynamicList = ({ list, deleteItem }) => {
+  const runTodoList = list.map(todo => (
+    <ContentDynamicTodo
+      key={todo.id}
```
Meno-101: Please have a look to the previous homework review.

homework-submission/react-hyf/react-hyf-client/src/React_HW3/index.jsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+
+class HomeworkWeek3 extends React.Component {
+  constructor() {
+    super();
+    this.state = {};
```
Meno-101: Initialing the state with no properties will be working perfectly in all case, but for a best practice and in case working in a strict mode applications, you don't have to initialize the state only, but with it's properties.

homework-submission/react-hyf/react-hyf-client/src/React_HW2/todoApp2.jsx
RamazanInce: I push the updated one for HW2

homework-submission/react-hyf/server/package.json
RamazanInce: Just delete from there will be enough?

homework-submission/react-hyf/react-hyf-client/src/React_HW3/index.jsx
RamazanInce:
```suggestion
   class HomeworkWeek3 extends React.Component {
  constructor(props) {
    super(props);
    this.state = {};
  }
```

RamazanInce: Will it be ok like that

homework-submission/react-hyf/react-hyf-client/src/React_HW2/todoApp2.jsx
RamazanInce: And implement it in HW3 also.

homework-submission/react-hyf/server/package.json
RamazanInce: We install those ones in the class and remain there.

RamazanInce: that is what the homework was asking may be I did it in a wrong place.

RamazanInce:
> Hello @RamazanInce,
> 
> Unfortunately I didn't find the connection between the server and the client

here what you mean 
const express = require('express');
const path = require('path');
const app = express();
const port = process.env.PORT || 8000;

app.use(express.static(path.join(__dirname, '../client/build')));

app.get('/', function(req, res) {
  res.sendFile(path.join(__dirname, '../client/build', 'index.html'));
});

app.listen(port, () => console.log(`Listening from server port ${port}`));

Meno-101:
Thanks for your reply,

I didn't say you haven't done it.
I said I haven't saw the connection between both:)

RamazanInce:
> Thanks for your reply,
> 
> I didn't say you haven't done it.
> I said I haven't saw the connection between both:)

It was different before..
now I am proposing the above one.
Is that ok

Meno-101:
Why would you assign a get method on the same rout that `use` uses?
keep the one you have it's the correct way.

Meno-101:
Waiting for the changes of the last week here :)

Meno-101:
Perfect, All working fine, Thanks for the update!

PR #444, opened 2019-07-19
Soolaf =>homework week4

soolaf:


homework-submission/myapp/src/App.js
```diff
@@ -0,0 +1,41 @@
...
+          <li>
+            <Link to='/week3'>Week3</Link>
+          </li>
+        </ul>
+
+        <div>
+          <Switch>
+            <Route exact path='/' render={() => <h3> Homeworks </h3>} />
+            <Route path='/week1' component={StaticList} />
+            <Route exact path='/week2' component={DynamicList} />
```
Meno-101: No need for an exact match here, it's only if you have a route that you could have another with the same name and different case, or you may have a root rout and you want to distinguish it from it's sub routes.

soolaf: Thank you @Meno-101 👍

PR #447, opened 2019-07-20
Enes w4

eneserdek:


homework-submission/react-hyf/server/server.js
```diff
@@ -0,0 +1,17 @@
+const express = require('express');
+
+const app = express();
+
+app.use(express.static('../react-hyf-client/build'));
+
+app.get('/', (req, res) => {
+  res.sendFile('../react-hyf-client/build/index', e => {
```
Meno-101: Oh no, I haven't notice this last time, this get method is the same as the `app.use` work.

PR #453, opened 2019-08-13
Suheyb - Week1

suh3yb:


homework-submission/main.js
```diff
@@ -0,0 +1,104 @@
...
+// and vice versa
+class ListItem extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { done: props.done ? props.done : false };
+    this.changeDoneState = this.changeDoneState.bind(this);
+  }
+
+  changeDoneState() {
+    if (this.state.done === true) {
```
ArcoMul: This code can be written shorter (and more readable). Hint: the if statement isn't needed

suh3yb:
Hi @ArcoMul, thank you so much for your feedback. 
I changed the code a bit to increase readability and to make it shorter and more organized. Also got rid of if statement as you suggested. 
Best regards

ArcoMul:
Super @suh3yb! You can write the if-statement which you now converted to a ternary even shorter:

```js
this.setState({ done: this.state.done ? false : true });
```

can be written as:

```js
this.setState({ done: !this.state.done });
```

suh3yb:
Oh this one is way better, thanks @ArcoMul :smile: I will change it right away.

PR #456, opened 2019-08-15
Static List and Dynamic List

ghifo:


ArcoMul:
Hi @ghifo!

Cool texture in the background of your page 😄 

A few things I noticed:
- You're using todoItems as a global variable in your DynamicList component. This is a bad practice. Please rewrite it in such a way that you pass the todos array as a prop to the component.
- You now have a prop called className. In my opinion this can be a bit confusing. It's better to just pass to the TodoItem if the todo item is done or not, and then decide in the TodoItem based on the done value what kind of class you add

Let me know here or on Slack if anything is not clear :-)

Good luck!

PR #457, opened 2019-08-16
hamza week-1

lizozo:


homework-submission/week-1/main.js
```diff
@@ -0,0 +1,82 @@
...
+    deadline: '2017-09-09',
+    done: false,
+  },
+];
+
+//with this component you can create a title for both lists by passing the title text as props
+class ListTitle extends React.Component {
+  render() {
+    const { text } = this.props;
+    return <h2 className={'list_title'}>{text}</h2>;
```
ArcoMul: No need to use className if the class is not dynamic

homework-submission/week-1/main.js
```diff
@@ -0,0 +1,82 @@
...
+    );
+  }
+}
+
+// this component creates a dynamic list for the todoItems object
+class DynamicList extends React.Component {
+  render() {
+    return (
+      <ul className="list">
+        {todoItems.map((todoItem, index) => {
```
ArcoMul: The homework isn't very explicit about it, but in my opinion it would be better to pass todoItems as a prop to this DynamicList component.

homework-submission/week-1/main.js
```diff
@@ -0,0 +1,82 @@
...
+          return <ListItem key={index} {...todoItem} />;
```
ArcoMul: No need to return in an arrow function. If you remove the curly brackets you don't have to explicitly return.

homework-submission/week-1/main.js
```diff
@@ -0,0 +1,82 @@
...
+
+//with this component you can create a title for both lists by passing the title text as props
+class ListTitle extends React.Component {
+  render() {
+    const { text } = this.props;
+    return <h2 className={'list_title'}>{text}</h2>;
+  }
+}
+
+// this component creates reusable list items for both lists
```
ArcoMul: Can you describe the different props and what their purpose is? Always good to write this down for other developers who have to work with your code.

lizozo:
Hi @ArcoMul !! Thanks for your valuable feedbacks!! I have made requested changes in "hamza changes for week1" commit. Kind regards

PR #458, opened 2019-08-16
wk1 static-dynamic-props

Nadoshka:


homework-submission/week1/app.js
```diff
@@ -0,0 +1,109 @@
...
+    const { title } = this.props;
+    return <h1>{title}</h1>;
+  }
+}
+
+// B- reusable li component
+//the goal of this component is to create a list item
+class Item extends React.Component {
+  render() {
+    // 4- Each list item should have a prop of description and deadline
```
ArcoMul: Great descriptions already. You're missing one prop here though, besides that it's good to describe the purpose of each prop. The `done` prop has an effect on what the component looks like, good to describe this here so that other developers know what happens when they use this component and its props.

homework-submission/week1/app.js
```diff
@@ -0,0 +1,109 @@
...
+
+// 3- Reuse one component at least 3 times
+// the goal of this component is to make static component which make render to the title list and three items
+class List extends React.Component {
+  render() {
+    return (
+      <div>
+        <ListTitle title="Static List" />
+        <ul>
+          <Item description={staticList[0].description} deadline={staticList[0].deadline} />
```
ArcoMul: If you use an array to define the static list it's better to loop over it in my opinion. The alternative is writing the data here statically/hard-coded.

homework-submission/week1/app.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+}
+
+// the goal of this component is to render the previous components (List, DynamicList)
+class App extends React.Component {
+  render() {
+    return (
+      <div>
+        <List />
+        <DynamicList items={secondTodoList} />
```
ArcoMul: Passing the todo items array as a prop, super!

week1/app.js
```diff
@@ -0,0 +1,204 @@
+var _createClass = function () { function defineProperties(target, props) { for (var i = 0; i < props.length; i++) { var descriptor = props[i]; descriptor.enumerable = descriptor.enumerable || false; descriptor.configurable = true; if ("value" in descriptor) descriptor.writable = true; Object.defineProperty(target, descriptor.key, descriptor); } } return function (Constructor, protoProps, staticProps) { if (protoProps) defineProperties(Constructor.prototype, protoProps); if (staticProps) defineProperties(Constructor, staticProps); return Constructor; }; }();
```
ArcoMul: What's the purpose of this file? :)

Nadoshka: I really don't know where this file comes from it seems like a compiled class maybe vsCode have added it somehow 😄 . I just removed it 👍

homework-submission/week1/app.js
Nadoshka: I agree 👍 , I understood that we are required to use the Item component three times. So that's why I have it here like this while in the dynamic list I'm looping over the items.

ArcoMul: Ha I see what you mean :-)

PR #459, opened 2019-08-16
Alusine-homework week1

alquinex:


week1/main.js
```diff
@@ -0,0 +1,65 @@
+// This is a class based component that has two props that are reuseable
+// as demonstrated in lines 47 through 49.
+class ListItems extends React.Component {
```
ArcoMul: `ListItem` would be a better name I think, since it's only one item of the list.

week1/main.js
```diff
@@ -0,0 +1,65 @@
...
+    );
+  }
+}
+
+// This component serves as parent component which renders the component
+// being nested in it.
+class App extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
```
ArcoMul: Good that you use state! Exactly what it is meant for.

week1/main.js
```diff
@@ -0,0 +1,65 @@
...
+          done: false,
+        },
+      ],
+    };
+  }
+  render() {
+    return (
+      <div className="list-items">
+        <h1 className="title">Static List</h1>
+        <ListItems description="Get out of bed" deadline="2017-09-13" />
```
ArcoMul: Please make a container component which holds the static list, and the same for the dynamic list. Two components called `StaticList` and `DynamicList` for example. Pass the right data to these container components and renders the `ListItem`s.

week1/main.js
alquinex: Hi @ArcoMul,
Thanks for your time and for the feedback. The changes have been made.

ArcoMul:
Hi @alquinex, the structure of your components is not very clear. You now have a component called StaticList, but it doesn't contain a list, it renders a list item. Your DynamicList uses a component called StaticList. A bit confusing.

Try to make your component structure like this:
```jsx
<div className="lists">
        <h1 className="title">Static List</h1>
        <StaticList /> { /* Renders your static list, using the same component for a list item as DynamicList */ }
        <h1 className="title">Dynamic List</h1>
        <DynamicList /> { /* Renders your dynamic list, using the same component for a list item as StaticList */ }
</div>
```

It's a bit hard to explain / demonstrate. See if you understand what I am saying. If not, talk to me on Slack then I can explain it more thoroughly, or we can make a call for a few minutes.

alquinex:
Hi @ArcoMul,
The changes have been made. Thanks for your guide -:)

PR #460, opened 2019-08-17
Hani's week 1 homework

hanialkudsi:


homework-submission/homework w1/app.js
```diff
@@ -0,0 +1,72 @@
+const ListItem = ({ description, deadLine, className }) => {
+  return <li className={className}>{description + ', ' + deadLine}</li>;
+};
+//ListItem is a reuseable component that has 3 props (description,deadline,className) and return a list item with description and deadLine separated by comma
+class StaticList extends React.Component {
+  render() {
+    return (
+      <ul>
+        Static list
```
ArcoMul: I don't think it is semantically correct html to place a string like this in your `<ul>`, better to place it above the `<ul>`

homework-submission/homework w1/app.js
```diff
@@ -0,0 +1,72 @@
...
+    done: false,
+  },
+];
+
+class DynamicList extends React.Component {
+  render() {
+    return (
+      <ul>
+        Dynamic list
+        {toDoList.map((item, i) => {
```
ArcoMul: It would be better to pass the toDoList as a prop to the component instead of using the global variable here.

homework-submission/homework w1/app.js
```diff
@@ -0,0 +1,72 @@
...
+          return (
+            <ListItem
+              description={item.description}
+              deadLine={item.deadline}
+              key={i}
+              className={item.done === true ? 'done' : 'not-done'}
```
ArcoMul: In my opinion passing className as a prop is a bit confusing. Better to pass the value of `item.done` as a prop and then in the `ListItem` component decide which class to add to the element.

homework-submission/homework w1/app.js
hanialkudsi: I don't really know either 
It's modified now

homework-submission/homework w1/app.js
hanialkudsi: yes i didn't really know how do it like this when i wrote the homework  but now i know thanks

ArcoMul:
Hi @hanialkudsi, 

Great work on fixing the `<ul>` and moving the todo items array to the state and pass it as a prop. 

Regarding passing the className to ListItem, you are still doing this, just by a different name ;-) The better way is to just pass the value of `item.done` to ListItem, and place the logic of which class to add to the element in ListItem

So instead of:
```jsx
<ListItem
  description={item.description}
  deadLine={item.deadline}
  key={i}
  done={item.done === true ? 'done' : 'not-done'}
/>
```
I'd like to see:
```jsx
<ListItem
  description={item.description}
  deadLine={item.deadline}
  key={i}
  done={item.done}
/>
```

If you have time, give it a try ✌️

PR #461, opened 2019-08-17
abdullah week1

heartwhite:
I was not present on Sunday. after Thursday I tried to figure out. 
I understood what I coded except for "super (props)"
I got an error about that and I sow it on homework of suheyb. 
I generally took help from @veli , and I looked on suheybs homework for concepts.


homework-submission/abdullahReact/main.js
```diff
@@ -0,0 +1,98 @@
...
+// in this case to set as done or not done by clicking
+
+class ListItem extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { done: props.done };
+    this.switchDoneState = this.switchDoneState.bind(this);
+  }
+
+  switchDoneState() {
```
ArcoMul: Cool that you can check and uncheck a todo item! Did you notice that when you click on an item in the static list that it doesn't get a line-through straight away? Do you know why?

homework-submission/abdullahReact/main.js
```diff
@@ -0,0 +1,98 @@
...
+  {
+    id: 3,
+    description: 'Eat breakfast',
+    deadline: '2017-09-09',
+    done: false,
+  },
+];
+
+//the goal of this component is to define/create a reusable list item that uses own function to set state.
+// in this case to set as done or not done by clicking
```
ArcoMul: Could you describe all the props this component can receive and what their goal is / what the effect is of the prop? This is very useful for other developers using your component.

homework-submission/abdullahReact/main.js
```diff
@@ -0,0 +1,98 @@
...
+      </ul>
+    );
+  }
+}
+//the goal of this component is to create an unordered list dynamically
+//it returns a ul which has items as many as given props.array has.
+class DynamicList extends React.Component {
+  render() {
+    const { toDoArr } = this.props;
+    console.log(this.props);
```
ArcoMul: Always good to remove `console.log`s before creating a pull request :)

homework-submission/abdullahReact/main.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+}
+
+//the goal of this component is to inject lists to html. it returns a div element which has two ul.
+class App extends React.Component {
+  render() {
+    return (
+      <div>
+        <StaticList />
+        <DynamicList toDoArr={toDoItems} />
```
ArcoMul: Great that you pass the todos array as a prop! 🎉

homework-submission/abdullahReact/package.json
```diff
@@ -0,0 +1,22 @@
+{
```
ArcoMul: What do you need the package.json for?

homework-submission/abdullahReact/main.js
heartwhite: thanks.

homework-submission/abdullahReact/main.js
heartwhite: yes. I forgot it, now I removed it.

homework-submission/abdullahReact/main.js
heartwhite: thanks for advise. I added more description

homework-submission/abdullahReact/main.js
heartwhite: I figured that out. 
when I call LitstItem inside Staticlist, I didn't pass the "done" property to write in state. 
that's why it tries to assign, but because of there is nothing it returns undefined. 
than "I guess" when I click, it takes the opposite of undefined it takes it like true. 
(by the way, the style "line-trough" applied when its false.) I corrected both the style and props.

homework-submission/abdullahReact/package.json
heartwhite: I don't know. I just write "nom init" automatically after NodeJS classes. 
I think I need it if I install React and ReactDOM modules and import right?

homework-submission/abdullahReact/main.js
ArcoMul: Ha super! Must have been quite a tricky one to figure out, I didn't see it on first sight. Cool! :-)

homework-submission/abdullahReact/main.js
ArcoMul: Great, totally fine like this. Just be aware that saying "// props.done is to initialize state.done." doesn't say a lot about the effect is of the prop. In my opinion a better comment would be: "// props.done gives the ListItem a line-through when it's true"

homework-submission/abdullahReact/package.json
ArcoMul: In this case you are loading React and ReactDOM using script tags in index.html. They are loaded in your browser straight from the url given in the script tag. So your package.json or your modules aren't used at all. No need to run npm init or something the like in this case :-)

homework-submission/abdullahReact/main.js
heartwhite: yes it was hard to see. wee looked that with my friend Talip in class. he noticed that.

homework-submission/abdullahReact/package.json
heartwhite: understand. I just don't used to write code so much. I think didn't get the programming mind yet. but I will soon. thanks for help

homework-submission/abdullahReact/main.js
heartwhite: yes it makes sense. I corrected that line

PR #462, opened 2019-08-17
aalhommada-react-week1

aalhommada:


homework-submission/HomeWork/main.js
```diff
@@ -0,0 +1,73 @@
+// create the Header component
+class Header extends React.Component {
```
ArcoMul: Super to a reusable component for the header; this would be a perfect case for a functional (presentational) component though :)

homework-submission/HomeWork/main.js
```diff
@@ -0,0 +1,73 @@
...
+  render() {
+    return <h2 className="header">{this.props.title}</h2>;
+  }
+}
+
+// create the list items component
+class List extends React.Component {
```
ArcoMul: Clear list item component! Think another time of the name of the component though. The component is not a List in my opinion. Also could you describe the props this component receives? And for each prop what its purpose/effect is?

homework-submission/HomeWork/main.js
```diff
@@ -0,0 +1,73 @@
...
+      </div>
+    );
+  }
+
+}
+// create Static list items and Dynamic list items using the previous component ( Header , List)
+class App extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
```
ArcoMul: Great that you put your todos in the state!

homework-submission/HomeWork/main.js
```diff
@@ -0,0 +1,73 @@
...
+      <div>
+        <Header title="Static List" />
+        <List description="Get out of bed" deadline="Wed Sep 13 2017" />
+        <List description="Brush teeth" deadline="Thu Sep 14 2017" />
+        <List description="Eat breakfast" deadline="Fri Sep 15 2017" />
+
+        <Header title="Dynamic List" />
+        {this.state.todos.map(todo => {
+          return (
+            <div key={todo.id}>
```
ArcoMul: Why do you put a `<div>` around each `<List>` component?

homework-submission/HomeWork/main.js
aalhommada: Thanks a lot 👍

homework-submission/HomeWork/main.js
aalhommada: Thanks @ArcoMul for your notes , i modify the code according your advice .

homework-submission/HomeWork/main.js
aalhommada: i made it one <div> now around the TodoListItems

ArcoMul:
Hi @aalhommada,

The name `TodoListItems` is better, in my opinion it would be even better to call the component `TodoListItem` though (without the s). In the end it renders only one item of the list right?

Which brings me to my next point, (I didn't realize this in my previous review), for every list item you are rendering an `<ul>`. You can see it if you expect it in the inspector.

afbeelding

Semantically this is not correct. Ideally we'd like to render a `<ul>` with multiple `<li>`'s inside. Can you see how you could solve this?

Let me know here or on Slack if it's not clear or if you are stuck :-)

aalhommada:
Thanks @ArcoMul  for the notes , i adjusted the  ``` TodoListItems ``` to  ``` TodoListItem ``` without s
and about the ``` ul ``` i made it on ul ,  indeed i thought that render a child element  like ``` li ``` in the component will not work  , and i should use parents like  ``` ul ```  or ``` div ```  but it is working , is my work now correct ?

ArcoMul:
Superrr! 🙌

PR #463, opened 2019-08-17
ghufran/week1

ghifo:


homework-submission/homework_week1/index.js
```diff
@@ -0,0 +1,80 @@
...
+  return (
+    <div className="dynamic_list">
+      <h1>Dynamic List</h1>
+      <ul>
+        {todoItems.map(({ id, description, deadline, done }) => (
+          <TodoItem
+            key={id}
+            description={description}
+            deadline={deadline}
+            className={done ? 'done' : ''}
```
ArcoMul: Your `<TodoItem>` component now has a prop called className, it's not very neat to pass the className as a prop. Better is to pass the value of done to `<TodoItem>`, and then in the TodoItem component you can decide based on this done prop which className to add.

homework-submission/homework_week1/index.js
```diff
@@ -0,0 +1,80 @@
+// Create a todo list item as reusable component:
```
ArcoMul: Could you describe the props which the component accepts? What is the goal/effect of the prop?

homework-submission/homework_week1/index.js
ghifo: Sure

ArcoMul:
Great @ghifo! The description of your component is a bit minimal still. It doesn't really describe the _effect_ of the props. Something like this would be better in my opinion:

```js
// Creates a todo list item
// Accepts the following props:
// - description: the content of the todo item
// - deadline: the date before which the todo item has to be done
// - done: if the todo item is done or not, add a line-through if true
class TodoItem extends React.Component {
...
```

ghifo:
Thanks for the tip!

PR #467, opened 2019-08-21
Suheyb - Week2

suh3yb:


homework-submission/main.js
```diff
@@ -0,0 +1,145 @@
...
+        {todoArr.map(todo => (
+          <ListItem key={todo.id} clickHandler={liClickHandler} todo={todo} />
+        ))}
+      </ul>
+    </section>
+  );
+};
+
+class Main extends React.Component {
+  state = { todoArr, behaviour: 'mark' };
```
shrynx: prefer passing the todoArr as a prop to the Main component rather than referencing a global variable

homework-submission/main.js
```diff
@@ -0,0 +1,145 @@
...
+    event.preventDefault();
+    const currentTodos = [...this.state.todoArr];
+    const newTodo = {
+      id: currentTodos[currentTodos.length - 1] ? currentTodos[currentTodos.length - 1].id + 1 : 1,
+      description: event.target.description.value,
+      deadline: event.target.deadline.value,
+      done: event.target.done.checked,
+    };
+    currentTodos.push(newTodo);
+    this.setState({ todoArr: [...currentTodos], behaviour: this.state.behaviour });
```
shrynx: tip, you don't need to give the total state to setState , only the state you want to change.
setState automatically takes rest of the values from the current state
https://reactjs.org/docs/state-and-lifecycle.html#state-updates-are-merged

homework-submission/main.js
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  markOrDeleteTodo(event) {
```
shrynx: how about splitting this into two functions ?
```
markOrDeleteTodo(event) {
  if (this.state.behaviour === 'mark') {
    markTodo()
  } else {
    deleteTodo()
  }
}
```

homework-submission/main.js
```diff
@@ -0,0 +1,145 @@
...
+    if (this.state.behaviour === 'mark') {
+      const copyArr = [...this.state.todoArr];
+      copyArr.forEach(todo => {
```
shrynx: instead of for each would it be better to use map ?

suh3yb:
Thank you very much @shrynx for your time and feedback. I have changed all parts of the code that you have pointed out.
Best regards.

homework-submission/main.js
```diff
@@ -85,7 +85,7 @@ const TodoList = ({ todoArr, liClickHandler, buttonClickHandler, clickBehaviour
 };
 
 class Main extends React.Component {
-  state = { todoArr, behaviour: 'mark' };
+  state = { todoArr: [...todoArr], behaviour: 'mark' };
```
ArcoMul: You are still using `todoArr` here in your Main component, even though this is a global variable. What @shrynx meant is to rewrite your code in such a way that you can render your Main component the following way:
```jsx
ReactDOM.render(<Main todoArr={todoArr} />, root);
```

homework-submission/main.js
```diff
@@ -0,0 +1,147 @@
...
+    event.preventDefault();
+    const currentTodos = [...this.state.todoArr];
+    const newTodo = {
+      id: currentTodos[currentTodos.length - 1] ? currentTodos[currentTodos.length - 1].id + 1 : 1,
+      description: event.target.description.value,
+      deadline: event.target.deadline.value,
+      done: event.target.done.checked,
+    };
+    currentTodos.push(newTodo);
+    this.setState({ todoArr: [...currentTodos] });
```
ArcoMul: You are already using the spread operator (`...`) and the new array operator (`[]`) to create a copy of the state. There is no need to do this again here on line 100 :-) This just results in unnecessary operations/performance/memory issues.

homework-submission/main.js
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  markTodo() {
+    const copyArr = this.state.todoArr.map(todo => {
+      if (todo.id.toString() === event.target.id) {
+        todo.done = !todo.done;
+      }
+      return todo;
+    });
+    this.setState({ todoArr: [...copyArr] });
```
ArcoMul: Same here, `Array.map` already creates a copy of the array in the state.

homework-submission/main.js
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  deleteTodo() {
+    const remainingTodos = this.state.todoArr.filter(
```
ArcoMul: And same for filter, also creates a copy of the array in the state.

suh3yb:
Thank you @ArcoMul for your time and feedback. Also thanks for your detailed explanations. I hope this time I would not miss any points :smile: 
Best regards.

fedefu:
Thanks for checking the homework @ArcoMul! Can we consider this PR approved?

Cheers!

ArcoMul:
Yes, great work @suh3yb!

suh3yb:
Thank you very much :smile:

PR #468, opened 2019-08-22
Heartwhite week2

heartwhite:


homework-submission/abdullahReact/main.js
```diff
@@ -0,0 +1,176 @@
...
+
+const Label = ({ text, className, htmlFor }) => {
+  return (
+    <label className={className} htmlFor={htmlFor}>
+      {text}
+    </label>
+  );
+};
+
+const Input = ({ name, className, type, placeholder, defaultValue }) => {
```
ArcoMul: Good idea to encapsulate default html elements as your own component, in that way you can always easily add custom behaviour through-out your whole app.

homework-submission/abdullahReact/main.js
```diff
@@ -0,0 +1,176 @@
...
+};
+
+const Form = ({ addListItem }) => {
+  const x = new Date();
+  const y = x.getFullYear().toString();
+  let m = (x.getMonth() + 1).toString();
+  let d = x.getDate().toString();
+  d.length == 1 && (d = '0' + d);
+  m.length == 1 && (m = '0' + m);
+  let date = y + '-' + m + '-' + d;
```
ArcoMul: You could move this date generation code to a separate function. So that here you would only write:
```js
let date = getDefaultDate();
```

homework-submission/abdullahReact/main.js
```diff
@@ -0,0 +1,176 @@
...
+      <ul>
+        <Header className="listHeader" text="Todo List" />
+        {toDoArr.map((item, i) => (
+          <ListItem
+            key={item.id}
+            id={item.id}
+            done={item.done}
+            description={item.description}
+            deadLine={item.deadLine}
+            onClick={deleteSelectedItem}
```
ArcoMul: I would call the `onClick` property something like `onDeleteClick` so that it's clear what the click event is linked to (a specific delete button)

homework-submission/abdullahReact/main.js
```diff
@@ -0,0 +1,176 @@
...
+    }
+    const newData = this.state.data;
+    const copyNewData = [...newData];
+    let deadLine = event.target.deadLine.value;
+    const dateTest = /([12]\d{3}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01]))/;
+    if (!dateTest.test(deadLine)) {
+      deadLine = 'not defined';
+    }
+    copyNewData.push({
+      id: copyNewData[copyNewData.length - 1].id + 1,
```
ArcoMul: This line is causing an error when I delete all todos in the list and then try to add one. Could you look into how to fix that?

homework-submission/abdullahReact/main.js
heartwhite: thanks.

homework-submission/abdullahReact/main.js
heartwhite: yes it makes sense. I changed my code. now default date is a function

homework-submission/abdullahReact/main.js
heartwhite: thanks for advice. now it is more readable. I changed that.

homework-submission/abdullahReact/main.js
heartwhite: yes. I declared a separate variable "id". I defined id = 1 with an if condition controlling length of items from state. If there are some items, then it takes from line. 
corrected, 
thank you for your time

fedefu:
Thanks for correcting this PR @ArcoMul! Can we consider it now approved?

ArcoMul:
Yes, well done @heartwhite! 👏

PR #469, opened 2019-08-23
7alip | Week2

7alip:


homework-submission/week2/main.js
```diff
@@ -0,0 +1,171 @@
...
+    };
+  }
+
+  handleAdd(description, deadline) {
+    // I any field is empty, alert user without adding the task
+    if (description === '' || deadline === '') return alert('Please fill all the fields!');
+
+    const tasks = [...this.state.tasks];
+    let _lastId = this.state.lastId;
+    tasks.push({
```
shrynx: although this is completely fine
prefer not mutating (pushing to an array) but make a new array

 const tasks = [...this.state.tasks, {
     id: lastId + 1,
      description,
      deadline,
      done: false
}];

homework-submission/week2/main.js
```diff
@@ -0,0 +1,171 @@
...
+  );
+};
+
+// Task container component
+class TaskContainer extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      tasks: this.props.data,
+      lastId: this.props.data.length
```
shrynx: since lastId can be calculated from tasks (as you already do), so this really doesn't need to be in the state.
whenever you need it just calculate the length of state.tasks

7alip: Yes, it's true. Here I just wanted for all current and deleted previous todos to have a unique ID. Just a fake `AUTO_INCREMENT` function which exist in databases 🤷‍♂️

7alip:
Hi @shrynx

Thank you so much for your nice feedback. I made some implementation after your feedback which is about controlled components although it is tedious as React Docs says 😫. It might be better to use Formik for next homeworks 🙂. 

I stored the states of input values in FormComponent but not in parent AppComponent. I'm not sure if it's the correct way instead of lifting state via props.

Best regards.

PR #474, opened 2019-08-26
Week2

hanialkudsi:


homework-submission/homework w1/app.js
```diff
@@ -0,0 +1,114 @@
+const ListItem = ({ description, deadLine, done }) => {
+  return <li className={done}>{description + ', ' + deadLine}</li>;
+};
+
+const Button = ({ text, type, event }) => {
```
shrynx: a better name would be for the the prop would be `clickHandler` or simply `onClick`.
as this function **onClick** takes an *event* and does something with it.
so onClick would be something like
```
onClick={event => doSomethingWith(event)}
```

homework-submission/homework w1/app.js
```diff
@@ -0,0 +1,114 @@
...
+
+  addToList(event) {
+    event.preventDefault();
+    const newTodoList = {
+      description: this.state.description,
+      deadLine: this.state.deadLine,
+      done: false,
+    };
+    let array = this.state.toDoList;
+    array.push(newTodoList);
```
shrynx: instead of `array.push` you can create a new array
```
const todoList = [...this.state.toDoList, newTodoList]
```

hanialkudsi: this way is much better thanks a lot!

homework-submission/homework w1/app.js
hanialkudsi: ok I think so I'll use it from now on

PR #475, opened 2019-08-27
Alusine-homework-week2

alquinex:


week2/main.js
```diff
@@ -0,0 +1,130 @@
...
+class AddItem extends React.Component {
+  handleChange = e => {
+    this.setState({
+      value: e.target.value,
+    });
+  };
+  handleSubmit = e => {
+    e.preventDefault();
+    this.props.onSubmit(this.state.value);
+    e.target.taskInput.value = '';
```
shrynx: rather than mutating the dom property directly, how about using set state here again ?

week2/main.js
```diff
@@ -0,0 +1,130 @@
...
+  }
+  addItem = (item, deadline) => {
+    const newItem = {
+      id: this.state.todos.length + 1,
+      description: item,
+      deadline,
+      done: false,
+    };
+    this.setState({
+      todos: [...this.state.todos, newItem],
```
shrynx: Nice 🙂

shrynx:
Hi @alquinex 
overall the code is nice. Good job !

i left one comment, and also one general advice, if your react component doesn't use state (or any lifecycles) then prefer using a function component rather than a class based.

alquinex:
Hi @shrynx 
Thanks for the feedback and for your time. All recommended changes have been made.

PR #476, opened 2019-08-30
hamza week-2

lizozo:


homework-submission/week-2/main.js
```diff
@@ -0,0 +1,121 @@
...
+        done: false,
+      },
+    ],
+  };
+  addTodo() {
+    event.preventDefault();
+    //If the state has 0 item, newId is 1, else id of the last element plus 1
+    const newId = this.state.todoItems.length
+      ? this.state.todoItems[this.state.todoItems.length - 1].id + 1
+      : 1;
```
shrynx: Nice 😃

homework-submission/week-2/main.js
```diff
@@ -0,0 +1,121 @@
...
+      {
+        id: 3,
+        description: 'Eat breakfast',
+        deadline: '2017-09-09',
+        done: false,
+      },
+    ],
+  };
+  addTodo() {
+    event.preventDefault();
```
shrynx: you are not passing the event argument to addTodo, so where does it appear from  ?🧐

shrynx:
overall this is pretty nice 🎉
i left a comment 
and also one general advice, if your component doesn't use state or lifecycles then prefer a function component instead of a class based component .
As a rule of thumb i always start with functional component and only promote it to class based component once i need state of life cycles

lizozo:
Hello @shrynx !! Thanks for the comments and review. Yes my methods do not pass "event" argument. I asked the same question in the class. I dont understand how it works this way. The weird thing is when i pass "event" as argument , my code doesnt work.

ArcoMul:
This is a bit magical to me too. Apparantly React or JS in general provides you by default with that event variable? Where does it come from?

It does make sense that you code doesn't work when you define it as a parameter on your function. If you update your `clickHandler` like this it would work:
```jsx
<InputForm clickHandler={event => this.addTodo(event)} />
```

PR #479, opened 2019-09-06
Abdullah week3 homework

heartwhite:


react-hyf/react-hyf-client/src/components/week3-homework.js
```diff
@@ -0,0 +1,68 @@
...
+      );
+  }
+  render() {
+    const { error, loaded, profiles, currentProfile } = this.state;
+    if (error) {
+      return <div>Error: {error.message}</div>;
+    } else if (!loaded) {
+      return <div>Loading...</div>;
+    } else {
+      console.log(this.state.profiles);
```
ArcoMul: Make sure to remove console.logs when creating a pull request ;-)

react-hyf/react-hyf-client/src/components/week3-homework.js
```diff
@@ -0,0 +1,68 @@
...
+      return (
+        <div className="week3-homework">
+          {loaded && <ProfileImage url={profiles[currentProfile].photo} />}
```
ArcoMul: Above you are already checking if `loaded` is false or true. When this code is being executed `loaded` is always true, so no need to check that again here right?

react-hyf/react-hyf-client/src/components/previous-homeworks.js
```diff
@@ -0,0 +1,37 @@
...
+  };
+  render() {
+    const { display, text } = this.state;
+    return (
+      <div>
+        <div className={display}>
+          <PreviousHomework1 />
+          <PreviousHomework2 />
+        </div>
+        <button className="show-hide-button" onClick={this.changeClass}>
```
ArcoMul: Cool idea to hide the previous homework behind a button :-)

react-hyf/react-hyf-client/src/components/week3-homework.js
heartwhite: I checked a few times. it is the last one I put again for another check. and I forgot. 
now removed.

react-hyf/react-hyf-client/src/components/week3-homework.js
heartwhite: actually, I don't know at all how it works.  I mean the "&&" means what or how it works. my friend Talip (7alip > GitHub) showed that to me when I share my problem.
now I tried, it works with only first row. I removed other two check.
thanks

PR #482, opened 2019-09-08
Alusine-homework-week3

alquinex:
Homework is incomplete, still working on it.


react-hyf/react-hyf-client/src/HomeworkWeek3.js
```diff
@@ -0,0 +1,48 @@
...
+
+  render() {
+    const { error, isLoading } = this.state;
+
+    if (error) {
+      return <h2 className="error">Oops!! There's a problem with the network.</h2>;
+    }
+
+    if (isLoading) {
+      return <Spinner />;
```
ArcoMul: Cool to get a complete random person from the API and show a spinner each time you are fetching!

react-hyf/react-hyf-client/src/components/HomeworkWeek3.js
```diff
@@ -0,0 +1,48 @@
+import React, { Component } from 'react';
```
ArcoMul: You have twice the same file/component in your repo (HomeworkWeek3), could you remove the one which is not being used?

PR #483, opened 2019-09-08
Week 3

lizozo:
I am sorry for the mess. I had a problem with pull request. please take the "hamza week-3" commit into consideration.

### DIFF
--- a/react-hyf/react-hyf-client/src/components/Homework3.js
+++ b/react-hyf/react-hyf-client/src/components/Homework3.js
@@ -1,0 +1,24 @@
+import React from 'react';
+import PhotoContainer from './PhotoContainer';
+import NameHeader from './NameHeader';
+import List from './List';
+class Homework3 extends React.Component {
+  render() {
+    const { name, surname, photo, gender, region, birthday, email, password, phone } = this.props;
+    return (
+      <div className='homework3'>
+        <NameHeader name={name} surname={surname}></NameHeader>
+        <PhotoContainer photo={photo}></PhotoContainer>
+        <List
+          gender={gender}
+          region={region}
+          phone={phone}
+          birthday={birthday.mdy}
+          email={email}
+          password={password}
+        ></List>
+      </div>
+    );
+  }
+}
+export default Homework3;
--- a/react-hyf/react-hyf-client/src/components/List.js
+++ b/react-hyf/react-hyf-client/src/components/List.js
@@ -1,0 +1,20 @@
+import React from 'react';
+import ListItem from './ListItem';
+
+class List extends React.Component {
+  render() {
+    const { gender, region, phone, birthday, email, password } = this.props;
+
+    return (
+      <ul className='listHomework3'>
+        <ListItem text={'👤' + gender} />
+        <ListItem text={'🌐' + region} />
+        <ListItem text={'📞' + phone} />
+        {<ListItem text={'🎂' + birthday.mdy} />}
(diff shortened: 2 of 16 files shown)


react-hyf/react-hyf-client/src/components/Homework3.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import PhotoContainer from './PhotoContainer';
+import NameHeader from './NameHeader';
+import List from './List';
+class Homework3 extends React.Component {
+  render() {
+    const { person } = this.props;
+    return (
+      <div className="homework3">
+        <NameHeader {...person}></NameHeader>
```
ArcoMul: Better to explicitly say which properties you are sending to the component, this way the code can become easily a bit unclear.

react-hyf/react-hyf-client/src/components/List.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import ListItem from './ListItem';
+
+class List extends React.Component {
+  render() {
+    const { gender, region, phone, birthday, email, password } = this.props;
+    console.log(birthday);
```
ArcoMul: Don't forget to remove `console.log`s

reacthyf/react-hyf-client/build/asset-manifest.json
```diff
@@ -0,0 +1,15 @@
+{
```
ArcoMul: It seems like your actual homework is in the folder called `react-hyf`, but you also have this folder `reacthyf`, I think it can be removed right?

homework-submission/week-1/index.html
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
```
ArcoMul: Your homework of week 1 and week 2 is also in here, could you remove that?

fedefu:
@lizozo can you please update your code so we close this PR? Cheers!

lizozo:
Hi @ArcoMul !! thanks for the feedback. I have made the changes about the code. But I am a little bit scared about  removing unnecessary commits and files. at the beginning I was unable to make a pull request becuz github said "commit histories are totally different and there is nothing to compare". So the mess was due to fix this issue. I need to work on git more. you may find the changes in "changes after feedback" commit. Have a nice week

ArcoMul:
Hi @lizozo, the point of Git is that you should feel free to delete files and code however you feel like. You can always rollback a commit/go back to a previous version in case something suddenly isn't working any more. If you are still afraid you might break something you can also make a copy of your whole project folder as a backup, so in case something goes wrong, you can restore it from that other folder (even I do this sometimes when I have to do very scare Git stuff ;-))

Your other changes are correct, but do please remove those other files. It's important you feel comfortable to do big changes sometimes in your code. In real life applications you also sometimes have to make big (scary) changes. Git is there to help you in those cases.

Good luck!

lizozo:
Hello @ArcoMul !! I tried to clean up the Pull Request. I removed everything except for the last version of the homework. Thank you very much. Have a nice week.

ArcoMul:
Hi @lizozo, super, all the right files are in there now.

Be aware that force pushing in general is a very bad idea. In theory you never need to force push something. Also in this case I can't imagine that it was really needed. If it happens that you can't push your code any more, then there is always the option to clone the branch again (cleanly), redo your changes, and then you should be able to push it from that new clone.

Anyway, happy to sign this one off :-)

PR #485, opened 2019-09-09
Suheyb - Week3

suh3yb:


react-hyf/react-hyf-client/src/homeworkWeek3.js
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }
+
+  async selectPerson() {
+    const { people } = this.state;
+    if (this.state.index > people.length - 1) {
+      await this.setState({ index: 0 });
+    }
+    const person = people[this.state.index];
+    const personArr = [
```
ArcoMul: It's better to save the `person` object data in the state, and then in the List component convert the data to an array to render it. That way the state data is better reusable for other component which do need the original object format. Now all components would have to depend on this array, which would be unclear.

react-hyf/react-hyf-client/src/homeworkWeek3.js
```diff
@@ -0,0 +1,109 @@
...
+      person.gender,
+      person.region,
+      person.phone,
+      person.birthday.dmy,
+      person.email,
+      person.password,
+    ];
+    this.setState({
+      personName: `${person.name} ${person.surname}`,
+      personInfo: [...personArr],
```
ArcoMul: `personArr` is a new array, no need to convert it to another new array in that case

react-hyf/react-hyf-client/src/homeworkWeek3.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  render() {
+    return (
+      <div className="App">
+        {this.state.error ? (
+          <Header className={`${this.state.hidden} error`} text="An error occurred. Try refresh." />
+        ) : (
+          <main className="main">
+            <Header className={`${this.state.hidden}`} />
```
ArcoMul: The `Header` component here receives a className called `true` or `false`, I don't really understand why.

suh3yb:
Hi @ArcoMul, thank you for your time and feedback. I changed the code as you suggested. 
Best regards.

ArcoMul:
Hi @suh3yb! I'm still a bit confused on how you use your `hidden` and `willHide` state properties, but the other changes are well done!

suh3yb:
Hi @ArcoMul. `hidden` and `willHide` state properties were confusing indeed, so I have gotten rid of them and refined the code. Instead of hiding with CSS, now I remove the whole component. I hope this time the code is better.
Thank you again for your time and feedback. :smile:

PR #487, opened 2019-09-10
Heartwhite week4

heartwhite:


react-hyf/server/server.js
```diff
@@ -0,0 +1,14 @@
+const Express = require('express');
```
shrynx: use lowercase for function names

PR #493, opened 2019-09-12
Alusine -homework-weeks4

alquinex:


alquinex:
Hi @shrynx,
Thanks for the feedback and for your time. Since we had to build the week4 homework based on the week3, I had to create week4 branch within week3. That's why you see a lot of files I guess.
I am actually having troubles now separating them.
But this is the path for the week4:
React/react-hyf/reactrouter
The week4 application and all its files sit in the 'reactrouter' folder.
I hope this will help ;)

fedefu:
@shrynx is this assignment clear or does it need to be pushed again? Thanks!

ArcoMul:
Hi @alquinex, I think all you have to do is to delete the folder `react-hyf-client`, since that doesn't seem to be used for the homework of week4.

Besides that would optionally be cool if `/` actually showed `/home`, a website should start at its homepage I would say :)

alquinex:
Hi @ArcoMul,
Looks better after the deletion, thanks! I thought of a different approach ;)

Hi @shrynx,
It looks better now with the PR;)

ArcoMul:
Super :-)

PR #496, opened 2019-09-14
Week4

lizozo:


lizozo:
> I am not sure what approach you took
> Maybe it was not clear, but the objective of this homework was to take individual component (not the whole apps) from previous homework and combine them in one app, which navigate to each component via ReactRouter

Hi @shrynx !! I am sorry for the mess. It is just "hamza week-4" commit for last week's homework. I had terrible moments with week-3 homework's pull request and it got very very messy. Please only consider the commit "hamza week-4" for the homework and i think it is what you are asking for. kind regards

shrynx:
@lizozo can you close this PR and make a new one with that commit only.
otherwise it is difficult to see

PR #504, opened 2019-10-03
Adham week1 react homework

Adham-Elias-Botours:
**This is ReactJs homework week 1.**

- homework as single html page: `homework-submission\week1\single_file_hw\week1_hw_single_file.html`

- homework as react app : `homework-submission\week1\homework_react_app`

thanks


homework-submission/week1/homework_react_app/src/components/list.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+// ListItem component: it is to create HTML list item.
+export const ListItem = props => {
+  return (
+    <li className={props.done ? 'line_through' : 'no_line'}>
+      {props.description} , {props.deadline}.
+    </li>
```
karamad10: this works fine, but maybe a little destructuring here would be cleaner.

karamad10:
Hey @Adham-Elias-Botours ,
well done! good use of various props, and the separate data file, only one small comment.
One thing you should also do is fix the missing key prop error you have.
And a little styling would be nice.

Adham-Elias-Botours:
**Are the adjustments good?**
Dear @ArcoMul  Thanks for your early review. 

**Questions:** 
1. Why there is no `key prop error` in the single HTML file? but just in the ReactApp?
2.  Is the current destructing for the `props` correct? , I preferred first to work with `props`, to understand the concept. How to destruct a thing which is not understandable?

I understand  form your review that I used `props` correctly, so for now If  the `destructing `  correct then I will be on the right way. But if not then I have to find more info about that.

For now styling is not in my priority, later there will be a time for it.

Thanks

karamad10:
Hey @Adham-Elias-Botours , 
Thank you for your reply.
the way you used props is 100% valid, but imagine having 100 props to use, would you want to type props. a 100 times? 
or having further nested props, so, for example, imagine the value you're looking for is 2 objects deep, would you rather do props.something.something.theNeededValue  
destructing is just a shorthand, it is not a concept on its own, but if you rather not to use it for now 
I don't see a problem.
As far as the key error thing, I am not sure really, but it is probably not a good thing that it is not catching the error. maybe @ArcoMul or @nick-barth  have something to say about this.

PR #505, opened 2019-10-03
Submission Homework-1

Mehmet-Ergorgec:


homework-submission/week1/src/App.js
```diff
@@ -0,0 +1,50 @@
...
+// I created them in theirs Container/Parent Component.
+// passed all of them in theirs sub-Components in order to use them.
+
+// List component includes ListHeader and ListItem Components.
+
+function App() {
+  return (
+    <div className="App-header">
+      <div className="StaticList">
+        <ListHeader listName={'Static List'} />
```
ArcoMul: Nice to make special ListHeader component :-)

homework-submission/week1/src/App.js
```diff
@@ -0,0 +1,50 @@
...
+        <ul>
+          <ListItem description={'Get out of bed'} deadline={'Wed Sep 13 2017'} />
```
ArcoMul: When you are passing simple strings to components you don't need to wrap then in `{' ... '}`, you can just use quotes "", for example: `description="Get out of bed"`

homework-submission/week1/src/components/listBody/index.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import ListItem from '../items';
+
+function ListBody({ items }) {
```
ArcoMul: Not saying it is wrong, but I was a bit lost in all the components you made. I think this ListBody component for example isn't really necessary. It good to create many reusable components, but too many components can also make your code harder to read for other developers 😉

homework-submission/week1/src/App.js
```diff
@@ -0,0 +1,50 @@
...
+  {
+    id: '76385kjsng2397528',
+    description: 'Eat breakfast',
+    deadline: '2017-09-09',
+    done: false,
+  },
+];
+// I created List, ListHeader and ListItem Components.
+
+// I have  props : listName, description,deadline and done for both Static and Dynamic List.
```
ArcoMul: Better to move these comments, the documentation of your components to the code of your components itself. So that other developers reading you code get a better understanding of how that specific part works :-)

homework-submission/week1/src/components/listBody/index.jsx
Mehmet-Ergorgec: I understand what you  mean. Now we have just started to learn React I am always thinking about how many components should I have? :)
Since we have learned about Components, I tried to use them. But I thought that What if I want to change items dynamically? I think It is easy to do it with ListBody Component. So I created it. 👍

Mehmet-Ergorgec:
@ArcoMul Thank you for your valuable feedback. 👍

PR #509, opened 2019-10-04
Mustafa/Week1

mayqf:


week1/src/components/headercomponent.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+
+const ComponentHeader=(props)=>{
+        return (
+        <div>
+          <h2 className="Componentheader">{props.title} </h2>
+        </div>
+        )
+        }
```
nick-barth: some ugly formatting here, are you using prettier?

week1/src/components/staticlist.js
```diff
@@ -0,0 +1,15 @@
+import React, { Component } from 'react'
+
+class Staticlist extends Component {
+   
+    render() {
+        return (
+         <div className="staticlist">
+            <ul>
+                <li> {this.props.description}/<b>Deadline :</b> {this.props.deadline}</li>
```
nick-barth: Destructure!

week1/src/components/headercomponent.js
mayqf: my laptop computer broke down. I think there is no prettier new one.I will take care

PR #514, opened 2019-10-05
react-week01-final

Laurelinduilwen:
homework submission for week01

phew, at least managed to complete it before midnight this time...


homework-submission/react-homework-w01/src/components/todoitem-static/index.jsx
```diff
@@ -0,0 +1,19 @@
+import React, { Component } from 'react';
+
+//component for static list, I guess nothing much to see here :o
+
+class TodoListStatic extends Component {
```
ArcoMul: The name `TodoListStatic` suggests that this is a whole list, but actually it's only on list **item**. Try to find a better name :-)

homework-submission/react-homework-w01/src/dynamiclist.json
```diff
@@ -0,0 +1,20 @@
+[
```
ArcoMul: If you didn't end up using the json file, maybe remove it from your code. Can be confusing for people reading the code.

homework-submission/react-homework-w01/src/components/todoitem-dynamic/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+  but I failed miserably and gave up and went with this solution using props. I am bad :(
+
+
+
+*/
+
+class TodoListDynamic extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
```
ArcoMul: Very good to use state to save your list.

homework-submission/react-homework-w01/src/components/todoitem-static/index.jsx
Laurelinduilwen: fixed since I destroyed the separate components to get a single one :)

homework-submission/react-homework-w01/src/dynamiclist.json
Laurelinduilwen: Its gone :) I actually included it in hopes of fixing the commented out code later :/

Laurelinduilwen:
Thank you for the feedback :) As I have mentioned in the comments I was having problems with the class syntax so I wanted to challenge myself a bit with this homework via adding extra classes. The new version is using the same component for the lists :3 

The in-code comments were actually a requirement and I was also not pleased with forcing a comment for each of my files since they were pretty basic, it is gone now :)

`Write for each component a short summary of what it does and the role of the prop (in case there is one) in a JavaScript comment`

ArcoMul:
Super! 👏

PR #516, opened 2019-10-07
Week 1 Homework

gulyener:
Sorry about handing in late


homework-submission/src/App.js
```diff
@@ -0,0 +1,58 @@
...
+      {
+        id: 3,
+        description: 'Eat breakfast',
+        deadline: '2017-09-09',
+        done: false,
+      },
+    ],
+  };
+
+  markComplete = id => {
```
ArcoMul: Cool that you already implemented a click event!

homework-submission/src/components/Todos.js
```diff
@@ -0,0 +1,20 @@
+import React, { Component } from 'react';
+import TodoItem from './TodoItem';
+
+const StaticTodos = ({ description, deadline }) => {
```
ArcoMul: You're using the plural work StaticTodos here while it only renders one todo item, might be confusing

gulyener:
Thanks for the review @ArcoMul. I have added the TodoItem to Static List but I could not remove the markComplete function in StaticTodo because TodoItem uses that function I guess. I'm not sure I did it the right way though, is this still static?

ArcoMul:
Super! 👏

PR #522, opened 2019-10-11
Homework Week 2

gulyener:
This week I didn't continue with last week's homework because it was Class based and I wanted to use functional components in this one.


homework-submission/hw-todolist/src/components/TodoList/index.jsx
```diff
@@ -0,0 +1,36 @@
...
+            }
+          >
+            <td>{todo.description}</td>
+            <td>{todo.deadline}</td>
+            {'      '}
+            <td>
+              <button onClick={() => deleteTodo(todo.id)}>Delete todo</button>
+            </td>
+            <td>
+              <button onClick={() => markTodo(todo.id, todo.done)}>Mark todo</button>
```
ArcoMul: Would be cool if the text of the 'Mark todo' button is 'Mark done' at fist and changes to 'Mark todo' once a todo is done

homework-submission/hw-todolist/src/components/TodoList/index.jsx
```diff
@@ -0,0 +1,36 @@
...
+      {todos.map(todo => {
+        return (
+          <tr
+            style={
+              todo.done === true ? { textDecoration: 'line-through' } : { textDecoration: 'none' }
+            }
+          >
+            <td>{todo.description}</td>
+            <td>{todo.deadline}</td>
+            {'      '}
```
ArcoMul: Why are these spaces needed?

gulyener: You're right I deleted it:)

gulyener:
Thank you for the review @ArcoMul. About the delete function, I couldn't see the bug you mentioned but I changed the id's just in case. I don't know if it's correct now. I also changed TodoList component to TodoTable since it's a table. Should I create Todo Items in a separate component?

ArcoMul:
Great improvements! `uuid` did fix the bug. What happened is that some todos ended up having the same id because you used `todos.length`

PR #524, opened 2019-10-11
Mustafa/week2

mayqf:


homework/homework2/src/components/button.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+function Button({ type = 'submit', value }) {
+  return (
+    <div>
```
ArcoMul: No need to wrap this in a `<div>`

homework/homework2/src/components/input.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+function Input({ type = 'text', placeholder, value, name, handleChangeInput }) {
+  return (
+    <div>
```
ArcoMul: Same, no `<div>` needed

homework/homework2/src/App.js
```diff
@@ -0,0 +1,107 @@
...
+          <Input
+            placeholder="description"
+            value={form.description}
+            name="description"
+            handleChangeInput={handleChangeInput}/>
+          <Input
+            placeholder="dd-mm-yyyy"
+            value={form.deadline}
+            name="deadline"
+            handleChangeInput={handleChangeInput}/>
```
ArcoMul: Better to use an input with type date here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date

homework/homework2/src/App.js
```diff
@@ -0,0 +1,107 @@
...
+    event.preventDefault();
+    if (form.description&&form.deadline) {
+      setTodos([...todos, form]);
+      setForm({
+        id: todos.length+1,
+        description: '',
+        deadline: '',
+        done: false,
+      });
+    } else {window.alert('Please enter a todo');
```
ArcoMul: Would be good to give a specific error, if the todo description or todo deadline is missing

homework/homework2/src/App.js
```diff
@@ -0,0 +1,107 @@
...
+    }
+  };
+  const [showTodoList, setShowTodoList] = useState(false);
+  const [buttonText, setButtonText] = useState(false);
```
ArcoMul: These two state properties could be moved to the form state and/or combined into one value. The buttonText property has in the end the same functionality as showTodoList, you could pick the button text based on the showTodoList value.

homework/homework2/src/components/button.js
mayqf: Many thanks. I did

homework/homework2/src/App.js
mayqf: I did

homework/homework2/src/App.js
```diff
@@ -43,6 +43,7 @@ function App () {
   const handleDeleteTodo = (id) => {
     const newTodos = todos.filter(item=>item.id!==id);
     setTodos(newTodos);
+    console.log(id);
```
ArcoMul: Always remove `console.log`s before creating a pull request

homework/homework2/src/App.js
```diff
@@ -54,14 +55,16 @@ function App () {
         deadline: '',
         done: false,
       });
-    } else {window.alert('Please enter a todo');
+    } else {window.alert('Please enter a todo description and select a deadline');
```
ArcoMul: What I meant is to create multiple `window.alert`s, so that based on which field is missing you get a different error message.

homework/homework2/src/App.js
```diff
@@ -54,14 +55,16 @@ function App () {
...
   const [showTodoList, setShowTodoList] = useState(false);
-  const [buttonText, setButtonText] = useState(false);
 
   const toggleList = () => {
-    return setShowTodoList(!showTodoList), setButtonText(!buttonText);
+    return setShowTodoList(!showTodoList);
+  };
+  const onFocusInput = e => {
+    e.currentTarget.type = 'date';
   };
```
ArcoMul: Why don't you just implement the html date input field? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date

ArcoMul:
Hi @mayqf, did you find time to look at why sometimes multiple todos are getting deleted? (hint: it's because of the ids which aren't always unique) and to see if you can use a date input?

mayqf:
> Hi @mayqf, did you find time to look at why sometimes multiple todos are getting deleted? (hint: it's because of the ids which aren't always unique) and to see if you can use a date input?
> 
> If you are stuck on either one of those things, let me know what you are stuck on then I can help you out.

mayqf:
> > Hi @mayqf, did you find time to look at why sometimes multiple todos are getting deleted? (hint: it's because of the ids which aren't always unique) and to see if you can use a date input?
> > If you are stuck on either one of those things, let me know what you are stuck on then I can help you out.

Many thanks.What is the problem with ids? I deleted one by one  with click. And it worked. You mean input type=date?

ArcoMul:
Hi @mayqf, good idea to use uuid, that fixes the issue that multiple todos can accidentality get deleted.

Currently your application doesn't start though, I get the following errors:
```
Error: Cannot find module "./App"
```

```
Failed to compile.

Error in ./src/index.js
Module not found: ./App in C:\Users\arcom\Development\hyf\class22\react\mayqf\homework\homework2\src

 @ ./src/index.js 13:11-27
```

Please look into that :-)

About the date input, why are you giving the input a `type="date"` when clicking the input? Why not give it a `type="date"` when rendering the html?

Good luck, know to find me here or on Slack if you need help!

mayqf:
error is because of chunk/wrong adding files. I deleted some files from commit. Probably that is the reason

PR #529, opened 2019-10-13
Hnn/week2

hnnhack:


homework-submission/hw-week2/src/App.js
```diff
@@ -0,0 +1,111 @@
...
+    super(props);
+    this.state = {
+      todo: dynamicList,
+    };
+  }
+
+  handleSubmit = e => {
+    e.preventDefault();
+    if (!e.target.itemField.value) {
+      alert('Text field is empty, please enter todo');
```
ArcoMul: Adding a return statement after this alert would remove the need to make an `} else {` statement, with easier readable code:

```js
if (!e.target.itemField.value) {
   alert('...');
   return;
}
...
```

homework-submission/hw-week2/src/App.js
```diff
@@ -0,0 +1,111 @@
...
+    } else {
+      const id = this.state.todo.length;
```
ArcoMul: Using the length of the todo as an id is a bit dangerous, since you can have multiple of the same id values.

homework-submission/hw-week2/src/App.js
```diff
@@ -0,0 +1,111 @@
...
+      const object = {
```
ArcoMul: It's true that this is an `object`, but actually it's a `todo` right? Try to name your variables so that it is very clear what value they are holding.

PR #533, opened 2019-10-16
Homework Week3

gulyener:
https://hardcore-ardinghelli-1fdcb5.netlify.com/
I can't believe I could actually deploy it:D
I tried to add country options but if I choose a couple of countries in a row some crazy things happen and I get a CORS error:(


ArcoMul:
Hi @gulyener, nice work already, I noticed a few things:
- When I load the page (in Firefox) the page grows and shrinks every time the loader goes around, I guess there goes something wrong
- In my opinion the loader should only show when the app is actually fetching data (so after the user pressed space bar and/or choose a country)
- Currently the page scrolls down when pressing space bar (default browser behaviour), could you try to prevent that?
- When not picking a country it works quite well, when I pick a country though afterwards the app starts to do multiple fetches when pressing space bar. I think you are not correctly cleaning up your callbacks or (fetch) events. Remember the clean up part of the `useEffect` function?

Please look into these things, and don't hesitate to ask for help here or on Slack.

Good luck!

gulyener:
Thank you so much for the review @ArcoMul  
I couldn't see the first point you mentioned so I couldn't fix that?.
I tried to fix 2nd and 3rd point. 
I tried to use stopPropagation to prevent the default behavior of space bar but I couldn't make it work. So I used preventDefault instead. 
For the last point, I think I solved the multiple fetching problem. But now when you choose a country you first get a user from that country, after that you have to click the button to get a user from that specific country. Otherwise if you press space bar you get a random user again. I don't know how to fix this. I left it at that because I want to move on to homework week4.

ArcoMul:
Great work @gulyener! I don't see the first point either any more, I guess its one of those things which fixed itself magically.

PR #555, opened 2019-10-24
Homework week 4

gulyener:
https://determined-hamilton-605992.netlify.com/

Here is my new link. I tried to use nested routes for hw assignments which took me ages to do. In the end I gave up on following the tutorials and just went trial-error. It worked finally. My file organization probably sucks as well as my code. I need to do some refactoring? and cleaning up. I also tried to explore some Bootstrap with this. Still really sloppy but better than my styling:D


ArcoMul:
Hi @gulyener, very clear code and file structure in my opinion. I really like the clean design as well! Well done!

gulyener:
Thanks so much @ArcoMul for the review and for spending your precious time on our assignments. It means a lot to us.

PR #603, opened 2020-10-03
Update W1-homework.md

Spardha-hash:
I add the learning resouces in this file and add bracket in the return statement, i think it makes the code more clear.


robvk:
Cleanup before updating React module

PR #604, opened 2020-11-14
Update codealong part

DegirmenciOmer:
I believe this video better suits our learning objectives  by offering less unfamiliar concepts in comparison to the  current one. Plus, it sounded more interesting to me.

### DIFF
--- a/week2/MAKEME.md
+++ b/week2/MAKEME.md
@@ -64,7 +64,11 @@ In this exercise we'll be using the following API endpoint: `https://official-jo
 
 > Create a new GitHub repository for this project. It's a portfolio piece!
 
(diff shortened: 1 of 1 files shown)


robvk:
I agree that adding the data grabbing portion is nice and lines up better. It is unfortunate that he spends half the time on the css which doesn't really add to this module. The effect is nice, but I think the students should be focusing on the react side.

@shrynx @nick-barth, what do you guys think?

DegirmenciOmer:
> I agree that adding the data grabbing portion is nice and lines up better. It is unfortunate that he spends half the time on the css which doesn't really add to this module. The effect is nice, but I think the students should be focusing on the react side.
> 
> @shrynx @nick-barth, what do you guys think?

Actually I skipped the css part while doing the codealong, my point is, it covers the week's objectives

PR #650, opened 2021-10-29
Add useCopy instructions with extra food for thought at the end

crevulus:
Wasn't sure how specific to go, so made the first 4 steps purposely ambiguous. Can make them more direct if needed.
Also added extra questions, though we can remove these if we want the practice exercises to be purely coding and less q&a,

### DIFF
--- a/week3/practice-exercises/01-copy-to-clipboard/README.md
+++ b/week3/practice-exercises/01-copy-to-clipboard/README.md
@@ -1,0 +1,23 @@
+# Copy to clipboard
+
+Can you make a hook that takes a user input and copies it to their computer's clipboard?
+
+1. Create a new application via `create-react-app`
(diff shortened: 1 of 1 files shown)


crevulus:
Found some issues. fixing now.

crevulus:
Actually the instructions are fine. The issue I was having was being code sandbox's browser doesn't support using `navigator.clipboard.writeText()` method. But it works just fine on an actual browser. 

Here's an example with some extra stuff to show how custom hooks can be used with React standard hooks: https://codesandbox.io/s/young-sky-hjv1f?file=/src/hooks/useCopy.js (open in browser at https://hjv1f.csb.app/)

robvk (approved):
Nice! Thanks a lot!

PR #685, opened 2022-11-15
Heba salem react.js week1 homework

HebaMak:
eCommerce app link on Netlify:

https://incomparable-beignet-a0fda3.netlify.app/


HebaMak:
eCommerce app link on Netlify;

https://incomparable-beignet-a0fda3.netlify.app/

HebaMak:
eCommerce app link on Nitlify:

https://incomparable-beignet-a0fda3.netlify.app/

week1/project/ecommerce/src/components/Products.jsx
```diff
@@ -0,0 +1,27 @@
...
+  const {products} = useContext(provideContext)
+
+  return (
+    <>
+      <h1 className="title">Products</h1>
+      <Categories />
+      <div className='products_container'>
+        {
+          products.map((product, index) => (
+          <Product key={index} product={product}/>
```
hadeel43, line 19: Looks good, but I prefer to not use index as a key of array - it has negative performance issues, instead we should try to find some unique and permanent value for keys

https://robinpokorny.medium.com/index-as-a-key-is-an-anti-pattern-e0349aece318

week1/project/ecommerce/src/components/Categories.jsx
```diff
@@ -0,0 +1,18 @@
+import allCategories from "../fake-data/all-categories"
+
+import Category from "./Category"
+
+function Categories() {
+  
+  return (
+    <div className="categories_container"> 
+      {
+        allCategories.map((category, index) => (
```
hadeel43, line 10: Same for the use of index :)

week1/project/ecommerce/src/components/Products.jsx
HebaMak: done

week1/project/ecommerce/src/components/Categories.jsx
HebaMak: categories array, has no ids and we are allowed to edit it,
once we work with the API I will change it to the category id

PR #705, opened 2023-10-23
made a little change in the README.md file

Mrhayor:


Mrhayor:
Two that where added in the README.md file which i removed one

Mrhayor:
I removed one of the that words and added a comma in the README.md file

